PR #9, opened 2020-03-26
Update Format

AurevoirXavier:
- [x] Follow substrate format, change line width from 120 to 100.
- [x] Format `Cargo.toml`.
- [ ] Add node template


HackFisher:
I will merge this first to avoid large conflicts with #12

PR #12, opened 2020-03-26
Try merge ring with kton using AccountBalanceData 

HackFisher:
+ [x] Fix warning issues
+ [x] Check other merging challenges
+ [x] Remove Kton and rename Ring etc.
+ [x] rename TryDropKton
+ [x] Check syntax : ```account_data.mutate_free(free);```, Moved to separate #24


frame/balances/ring/src/lib.rs
```diff
@@ -291,9 +297,9 @@ decl_storage! {
 					"the balance of any account should always be more than existential deposit.",
 				)
 			}
-			for &(ref who, free_ring) in config.balances.iter() {
+			for &(ref who, free) in config.balances.iter() {
 				let mut account_data = T::AccountStore::get(who);
-				account_data.free_ring = free_ring;
+				account_data.mutate_free(free);
```
HackFisher: Review this: why compile succucess

HackFisher: ```^^^^^ cannot infer type for type parameter `T` declared on the trait `Data````

when multi trait impl

HackFisher: A simulation, it seems that this will works:

https://play.rust-lang.org/?version=stable&mode=debug&edition=2015&gist=3be99c17bcc5244d75507d273ac7ea67

HackFisher:
Can not pass the factory test on darwinia repo, so hold on.

clearloop:
> Can not pass the factory test on darwinia repo, so hold on.

I'm not familiar with the `balance` pallet, but I know the `treasury` pallet can not implement this because of the [`decl_module!`][0] macro doesn't support `<>` like parameters in methods.(if this blocks us, I would like to try to modify this macro)

I'll follow this issue/pr when I'm free with the works on my hand, interested in the `trait` and `macro` in rust.


[0]: https://github.com/paritytech/substrate/blob/9e1973967cf313f8f304217989ce94916ccdb967/frame/support/src/dispatch.rs#L228

AurevoirXavier:
> > Can not pass the factory test on darwinia repo, so hold on.
> 
> I'm not familiar with the `balance` pallet, but I know the `treasury` pallet can not implement this because of the [`decl_module!`](https://github.com/paritytech/substrate/blob/9e1973967cf313f8f304217989ce94916ccdb967/frame/support/src/dispatch.rs#L228) macro doesn't support `<>` like parameters in methods.(if this blocks us, I would like to try to modify this macro)
> 
> I'll follow this issue/pr when I'm free with the works on my hand, interested in the `trait` and `macro` in rust.

Not this problem and we don't want to maintain too many crates.

clearloop:
> Not this problem and we don't want to maintain too many crates.

Just pr substrate if okay.

PR #26, opened 2020-03-30
Patch Merge Ring and Kton

AurevoirXavier:


bin/node-template/node/src/chain_spec.rs
```diff
@@ -178,15 +179,14 @@ fn testnet_genesis(
 		pallet_claims: Some(Default::default()),
 		pallet_eth_backing: Some(Default::default()),
 		pallet_eth_relay: Some(Default::default()),
-		// pallet_eth_backing: Some(Default::default()),
-		pallet_balances_Instance1: Some(KtonConfig {
+		pallet_balances_Instance0: Some(RingConfig {
```
HackFisher: Instance is not start from 1? 

Are DefaultInstance different with Instance0?

And it should be ok to select Instance1 and Instance2, the node codes from substrate, collectives are using instance1 and instance2

AurevoirXavier: Yes, from 0 to 15.

cargo expand:
```rust
pub struct DefaultInstance;
pub struct Instance0;
...
pub struct Instance15;
```

PR #33, opened 2020-03-31
ci: use the latest nightly

### LINKED ISSUE #32: Set Rust Version to Latest `Nightly` in CI

yanganto:
We ran into a compile problem with nightly around Feb 2020, and the issue may be solved.
So we free the version limitation.
- free the nightly version
- fix #32


AurevoirXavier:
Also the darwinia repo.

PR #36, opened 2020-04-02
Refactor header mmr to use Other Digest Item Type

HackFisher:
Closes https://github.com/darwinia-network/darwinia/issues/410


frame/header-mmr/src/lib.rs
```diff
@@ -45,13 +45,32 @@
...
 use frame_support::{decl_error, decl_module, decl_storage, ensure};
-use sp_runtime::{generic::DigestItem, traits::Hash, DispatchError};
+use sp_runtime::{
+	generic::{DigestItem, OpaqueDigestItemId},
+	traits::{Hash, Header},
+	DispatchError, RuntimeDebug,
+};
 use sp_std::{marker::PhantomData, prelude::*};
 
+pub const MRR_ROOT_LOG_ID: [u8; 4] = *b"MMRR";
```
AurevoirXavier, line 63: What does the MRR means?

PR #38, opened 2020-04-03
Refactor eth network const and subkey prefix

### LINKED ISSUE #19: Refactoring Eth-Relay and Eth-Backing Tests
- [x] Testing canonical header re-org, related to darwinia-network/darwinia#302
- [x] Block Safe for verification receipt darwinia-network/darwinia#313 
- [x] Testing for RingReward / KtonReward, current reward OnUnbalance is set to (), and imblance when on_unbalance should be zero after redeeming, because balance are transfer to target redeem account

HackFisher:
Closes #19


AurevoirXavier:
We can check the prefix in `DetermineAccountId`'s impl. And I think this trait is useless (always use subkey in darwinia). It's troublesome, we've to define it in the runtime.

PR #42, opened 2020-04-07
Fix Bug

AurevoirXavier:


.maintain/types/types.json
```diff
@@ -1,4 +1,5 @@
 {
+	"Address": "AccountId",
```
HackFisher, line 2: Suggest to move types to runtime folder and keep it consistent with the type in the runtime.

For here in node-template, it is using ```pub type Address = <Indices as StaticLookup>::Source``` now.

PR #43, opened 2020-04-08
Get token from offchain db

yanganto:
- Get the token from offchain db


yanganto:
The ethscan apikey should be inserted with js function here
```js
api.rpc.offchain.localStorageSet(1,"eapi","XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX")
```
The interface of js client is confirmed in builders chanel.

So there is some issue that offchain worker can get the key, but not the value.
I have comfirm the rpc call, and I think there is a issue in `sp_runtime::offchain::storage::StorageValueRef`.

yanganto:
The token should be 34 chars.

PR #60, opened 2020-04-11
Support `EthHeader::from_str_unchecked`

AurevoirXavier:

PR #63, opened 2020-04-13
Multiple Ethereum resource

yanganto:
- redundant with multiple ethereum resource
- get block from ethscan if api key provided else from ethgateway
- https://github.com/darwinia-network/darwinia/issues/371


AurevoirXavier:
conflict with #60

pending

yanganto:
The post request of ether gateway follows the standard
https://github.com/ethereum/wiki/wiki/JSON-RPC#eth_getblockbynumber

PR #82, opened 2020-04-17
Storage Migration for eth-relay

AurevoirXavier:


primitives/eth-primitives/src/header.rs
```diff
@@ -609,82 +603,6 @@ mod tests {
 		);
 	}
 
-	#[test]
-	fn mix_hash_should_work_for_mainnet_block_0x1() {
```
HackFisher, line 613: Why remove this test?

AurevoirXavier: we don't use new dag anymore

and it takes too much time in tests

AurevoirXavier: move to feature

https://github.com/darwinia-network/darwinia-common/pull/82/commits/827e3c2320f1c8a9cffa9794ee7ad7c758b5b274

PR #95, opened 2020-04-23
Add offchain test

### LINKED ISSUE #69: Add more tests for eth-offchain

yanganto:
- Add more integration test for offchain
- Closes #69


HackFisher:
New pull request should based on master codebase, I don't think it's a good practice to based on another pull request, since not all the pull requests will be merged finally.

Don't need to worry about merge at the first commits, we can always try merge master at 
 final merge, at most we need to resolving some easy-to-fix conflicts.

This could free your a lot attention away from considering other pull requests during focusing on you pull request works.

For me, I haven't found conflicts hard to resolve for quite a long time, especially for small changes.

frame/bridge/eth/offchain/src/lib.rs
```diff
@@ -303,24 +318,33 @@ impl<T: Trait> Module<T> {
...
 	}
 
 	/// Extract the proof no mater the response is scale encoded format or json format
 	fn extract_proof(r: &mut Vec<u8>, option: bool) {
+		// Note the size is checked in `validate_response`, so there is no check here
+
 		let (hint, left_offset, right_offset) = if option { (125, 11, 5) } else { (44, 12, 3) };
-		let mut pr = 47;
-		for i in 47..r.len() {
+		let mut pr = 1355;
```
HackFisher, line 346: I think the numbers here are quite magic, need some comments on it. Or better to use robust utils/libraries instead.

frame/bridge/eth/offchain/src/lib.rs
```diff
@@ -334,43 +358,41 @@ impl<T: Trait> Module<T> {
...
-				u8::from_str_radix(from_utf8(&resp_body[i..i + 2]).unwrap_or_default(), 16)
-					.unwrap_or_default()
-			})
-			.collect();
-		let may_decoded_header: Option<EthHeader> =
-			Decode::decode::<&[u8]>(&mut &scale_bytes[..]).ok();
-		may_decoded_header.unwrap_or_default()
+		// Note the size is checked in `validate_response`, so there is no check here
+
+		let i = resp_body[50..]
```
HackFisher, line 390: Magic number here.

frame/bridge/eth/offchain/src/lib.rs
```diff
@@ -203,21 +213,20 @@ impl<T: Trait> Module<T> {
...
-				trace!(target: "eoc-rh", "[eth-offchain] request fail back wth option");
+				info!(target: "eoc-rh", "[eth-offchain] request without option fail: {:?}", e);
+				info!(target: "eoc-rh", "[eth-offchain] request fail back wth option");
 				Self::fetch_header(ETHRESOURCE.to_vec(), target_number, true)?
 			}
 		};
 
+		// The `submit_header` will call event out of eth-offchain pallet,
+		// so this function is skiped in the test of pallet
+		#[cfg(not(test))]
```
HackFisher, line 229: This looks tricky, should have better ways.

- we could put the tested code in the function to a seperate method.
- some other unit test might want to test the codes with ```submit_header```

frame/bridge/eth/offchain/src/lib.rs
yanganto: Fixed, thanks for review.

frame/bridge/eth/offchain/src/lib.rs
yanganto: Fixed.  Thanks.

frame/bridge/eth/offchain/src/lib.rs
yanganto: As discussed in zoom meeting, this is tricky but there is not realy big things need to be test in that function.  Please take a look and find out if it is acceptable.  Thanks.

PR #114, opened 2020-04-28
easy-testing for eth-primitives

AurevoirXavier:


frame/bridge/eth/offchain/src/lib.rs
```diff
@@ -261,7 +261,8 @@ impl<T: Trait> Module<T> {
 		let eth_header_part =
 			Self::extract_from_json_str(&resp_body[..], b"eth_header" as &[u8]).unwrap_or_default();
 		let header = if option {
-			EthHeader::from_str_unchecked(from_utf8(eth_header_part).unwrap_or_default())
+			panic!("FIXME")
```
AurevoirXavier, line 264: @yanganto

primitives/eth-primitives/src/header.rs
```diff
@@ -783,11 +804,44 @@ mod tests {
...
+		let decoded_header: EthHeader = Decode::decode::<&[u8]>(&mut scale_encoded).ok().unwrap();
```
HackFisher, line 809: Scale encoded have this linux and golang stuffs?

HackFisher: 3GdGeth/v1.0.0/linux/go1.4.2\xd6~ME\x03C\x04d%\xaeBqGCS\

AurevoirXavier: @yanganto

primitives/eth-primitives/src/header.rs
```diff
@@ -783,11 +804,44 @@ mod tests {
...
 		assert_eq!(header, decoded_header);
 	}
+
+	#[test]
+	fn deserialize_should_work() {
+		let header = EthHeader::from_str_unchecked(
+			r#"
+			{
+				"difficulty": "0x234ac172",
+				"extraData": "0xde830207028f5061726974792d457468657265756d86312e34312e30826c69",
```
HackFisher, line 819: field names should be updated according spec?
and from_str_unchecked can be simplified

AurevoirXavier: I prefre to keep this function. It can handle all raw response from etherscan ethgateway web3 API. Easy testing for eth-primitives.

AurevoirXavier:
> since pub fn from_str_unchecked(s: &str) is only used by mod test, suggest to move to mod test,
> 
> but it ok to keep as it is

Yes, I already mark it as `#[cfg(feature = "easy-testing")]`.

primitives/eth-primitives/src/header.rs
clearloop: Maybe `extra_data` with `geth version`

Issue #109, opened 2020-04-26
Fix warnings and Error logs in cargo test --release


HackFisher:
```
warning: unused import: `offchain::http::Request`
  --> frame/bridge/eth/offchain/src/lib.rs:53:18
   |
53 | use sp_runtime::{offchain::http::Request, traits::Zero, DispatchError, KeyTypeId};
   |                  ^^^^^^^^^^^^^^^^^^^^^^^
   |
   = note: `#[warn(unused_imports)]` on by default

warning: unused variable: `header`
   --> frame/bridge/eth/offchain/src/lib.rs:218:8
    |
218 |         let (header, proof_list) = match header_without_option {
    |              ^^^^^^ help: consider prefixing with an underscore: `_header`
    |
    = note: `#[warn(unused_variables)]` on by default

warning: unused variable: `proof_list`
   --> frame/bridge/eth/offchain/src/lib.rs:218:16
    |
218 |         let (header, proof_list) = match header_without_option {
    |                      ^^^^^^^^^^ help: consider prefixing with an underscore: `_proof_list`

warning: constant item is never used: `MAX_REDIRECT_TIMES`
  --> frame/bridge/eth/offchain/src/lib.rs:65:1
   |
65 | const MAX_REDIRECT_TIMES: u8 = 3;
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = note: `#[warn(dead_code)]` on by default

warning: field is never read: `location`
  --> frame/bridge/eth/offchain/src/lib.rs:72:2
   |
72 |     location: Vec<u8>,
   |     ^^^^^^^^^^^^^^^^^

warning: field is never read: `payload`
  --> frame/bridge/eth/offchain/src/lib.rs:73:2
   |
73 |     payload: Vec<u8>,
   |     ^^^^^^^^^^^^^^^^

warning: field is never read: `redirect_times`
  --> frame/bridge/eth/offchain/src/lib.rs:74:2
   |
74 |     redirect_times: u8,
   |     ^^^^^^^^^^^^^^^^^^

warning: field is never read: `cookie`
  --> frame/bridge/eth/offchain/src/lib.rs:75:2
   |
75 |     cookie: Option<Vec<u8>>,
   |     ^^^^^^^^^^^^^^^^^^^^^^^

warning: method is never used: `submit_header`
   --> frame/bridge/eth/offchain/src/lib.rs:309:2
    |
[...]

PR #128, opened 2020-05-07
Suppress warning in offchain test

yanganto:
- suppress dead code warning in offchain test
- relate to #109


HackFisher:
Suppress dead code warning in travis ci is not so important for #109, the intention of #109 is to clean and fix the warnings when run ```cargo test --release```.

Suppress warning in rust lang is not good practice IMO.

PR #144, opened 2020-05-15
Fix Bugs

AurevoirXavier:


frame/claims/src/lib.rs
```diff
@@ -189,13 +233,31 @@ decl_module! {
...
+						);
+						T::RingCurrency::transfer(
+							&Self::account_id(),
+							&dest,
+							balance_due,
+							KeepAlive,
+						)?;
+					}
+
+					#[cfg(not(feature = "init-supply"))]
```
HackFisher, line 251: I guess this should be directly removed? No need to maintain two versions.

frame/claims/src/lib.rs
```diff
@@ -204,13 +266,31 @@ decl_module! {
...
```
HackFisher, line 284: same here.

frame/claims/src/lib.rs
```diff
@@ -224,10 +304,16 @@ decl_module! {
...
 			match who {
 				OtherAddress::Eth(who) => {
+					#[cfg(feature = "init-supply")]
+					T::RingCurrency::deposit_creating(&Self::account_id(), value);
+					#[cfg(not(feature = "init-supply"))]
 					<Total<T>>::mutate(|t| *t += value);
 					<ClaimsFromEth<T>>::insert(who, value);
 				}
 				OtherAddress::Tron(who) => {
+					#[cfg(feature = "init-supply")]
```
HackFisher, line 314: same here.

frame/claims/src/lib.rs
```diff
@@ -161,7 +191,13 @@ decl_storage! {
 				}
 			}
 
+			#[cfg(not(feature = "init-supply"))]
 			<Total<T>>::put(total);
+			#[cfg(feature = "init-supply")]
+			let _ = T::RingCurrency::make_free_balance_be(
```
HackFisher, line 197: This should be called on migrate too.

frame/claims/src/lib.rs
```diff
@@ -7,7 +7,27 @@ mod address;
 #[cfg(feature = "std")]
 pub use address::*;
 
+#[cfg(feature = "init-supply")]
```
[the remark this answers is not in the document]
HackFisher: I prefer not to use feature here.

frame/claims/src/lib.rs
```diff
@@ -7,7 +7,27 @@ mod address;
...
+mod migration {
+	// --- substrate ---
+	use frame_support::migration::*;
+	// --- darwinia ---
+	use crate::*;
+
+	pub fn migrate<T: Trait>() {
+		sp_runtime::print("Migrating DarwiniaClaims...");
+
+		if let Some(_) = take_storage_value::<RingBalance<T>>(b"DarwiniaClaims", b"Total", &[]) {
```
HackFisher, line 20: Substrate's StorageVersion control practice can been seen as a good practice, though we may do not need to use it in the case here:

https://github.com/paritytech/substrate/commit/497e8a094129fefa052ffb43918de8de8fe66c29#diff-32ce5961608780655bf3e00b2cc07d59L1252

AurevoirXavier: Yes, I've already review that. But the staking module is so complex.

This situation is similar to our.

https://github.com/paritytech/substrate/pull/5010

frame/claims/src/lib.rs
```diff
@@ -7,7 +7,39 @@ mod address;
...
+	// --- darwinia ---
+	use crate::*;
+
+	pub fn migrate<T: Trait>() {
+		sp_runtime::print("Migrating DarwiniaClaims...");
+
+		if let Some(total) = take_storage_value::<RingBalance<T>>(b"DarwiniaClaims", b"Total", &[])
+		{
+			let minimum_balance = T::RingCurrency::minimum_balance();
+			let _ = T::RingCurrency::make_free_balance_be(
```
HackFisher, line 22: Not sure for this usage, because didn't found sample usage like module function call in Substrate codebase. 

Need confirming.

PR #160, opened 2020-06-10
Add static tests for header-mmr

clearloop:
# Desc
A static tests ref to the result of subscan, to make sure the result of :

+ New & push batch of header hashes into mmr at one time
+ New & push header hashes one by one

are the same.

### DIFF
--- a/frame/header-mmr/src/mock.rs
+++ b/frame/header-mmr/src/mock.rs
@@ -84,3 +84,59 @@ pub fn initialize_block(number: u64, parent_hash: H256) {
 		Default::default(),
 	);
 }
+
+// -- helpers ---
+pub const HEADERS_N_ROOTS: [(&str, &str); 10] = [
+	(
+		"34f61bfda344b3fad3c3e38832a91448b3c613b199eb23e5110a635d71c13c65",
+		"34f61bfda344b3fad3c3e38832a91448b3c613b199eb23e5110a635d71c13c65",
+	),
+	(
+		"70d641860d40937920de1eae29530cdc956be830f145128ebb2b496f151c1afb",
+		"3aafcc7fe12cb8fad62c261458f1c19dba0a3756647fa4e8bff6e248883938be",
+	),
+	(
+		"12e69454d992b9b1e00ea79a7fa1227c889c84d04b7cd47e37938d6f69ece45d",
+		"7ddf10d67045173e3a59efafb304495d9a7c84b84f0bc0235470a5345e32535d",
+	),
+	(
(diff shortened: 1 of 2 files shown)


frame/header-mmr/src/mock.rs
```diff
@@ -84,3 +84,59 @@ pub fn initialize_block(number: u64, parent_hash: H256) {
...
+		"620dbc3a28888da8b17ebf5b18dba53794621463e2bbabcf88b8cbc97508ab38",
+	),
+	(
+		"c0c8c3f7dc9cdfa87d2433bcd72a744d634524a5ff76e019e44ea450476bac99",
+		"a94bf2a4e0437c236c68675403d980697cf7c9b0f818a622cb40199db5e12cf8",
+	),
+];
+
+/// Convert hex string to H256
+pub fn hh(s: &str) -> H256 {
```
AurevoirXavier, line 133: use array_bytes::fixed_hex_bytes_unchecked;

frame/header-mmr/Cargo.toml
```diff
@@ -33,6 +35,8 @@ std = [
 	# crates
 	"codec/std",
 	"serde",
+    # darwinia
+	"array-bytes/std",
```
AurevoirXavier, line 39: remove this

frame/header-mmr/Cargo.toml
```diff
@@ -14,6 +14,8 @@ codec = { package = "parity-scale-codec", version = "1.3.0", default-features =
 serde = { version = "1.0.106", optional = true }
 # github
 merkle-mountain-range = { package = "ckb-merkle-mountain-range", default-features = false, git = "https://github.com/nervosnetwork/merkle-mountain-range.git" }
+# darwinia
+array-bytes = { default-features = false, path = "../../primitives/array-bytes" }
```
AurevoirXavier, line 18: only deps this in dev-dependencies.

frame/header-mmr/Cargo.toml
```diff
@@ -25,6 +23,8 @@ sp-std = { default-features = false, git = "https://github.com/darwinia-network/
 [dev-dependencies]
 # crates
 serde_json = { version = "1.0.51" }
+# darwinia
+array-bytes = { default-features = true, path = "../../primitives/array-bytes" }
```
AurevoirXavier, line 27: default-features can be ignore in dev-dependencies.

clearloop: ...

AurevoirXavier: Because default just mean default... Haha.

PR #167, opened 2020-06-30
Ethereum Relay

yanganto:
The api to store the brief of header in Relayable trait is missing.
The `last_confirmed` dependent on the header brief input from the `store_header_brief`.


AurevoirXavier:
It should be `store_header`.

We need the receipt for backing, brief doesn't contains those things.

yanganto:
It will be good to implement the API with a clear interface. 
It is not a huge trait we can well-define it before implementation.  
The API `store_header` looks good for me.

frame/bridge/ethereum/relay/src/lib.rs
```diff
@@ -1,67 +1,261 @@
...
+	///              /   \
+	///             -     3rd hash of mmr_proof
+	///           /   \
+	///          -     2nd hash of mmr_proof
+	///        /   \
+	/// leaf_hash   1st hash of mmr_proof
+	///
+	/// mmr_proof = [ 1st_hash, 2nd_hash, 3rd_hash ... nth_hash ]
+	///
+	fn verify_mmr(
```
HackFisher, line 217: Need to check the spec of the MMRProof here with the proof generated by nervos lib.

HackFisher: Implementation need to use nervos lib's verify:

https://github.com/nervosnetwork/merkle-mountain-range/blob/ea232474496bdf072d7de33297fd54f1ef8df599/src/mmr.rs#L271

yanganto: The two functions are used to generate MerkleProof and verify the proof within nervos lib.
```pub fn new(mmr_size: u64, proof: Vec<T>) -> MerkleProof ```
```pub fn verify(&self, root: T, leaves: Vec<(u64, T)>) -> Result<bool> ```

Following parameters are needed from relayer.
1. proof
2. all leaves with positions
3. mmr root

`mmr_size` is assumed be calculated as mentioned in https://github.com/darwinia-network/darwinia-common/issues/177,
if not  the mmr_size also need to be provided by relayer.

File: frame/bridge/ethereum/relay/src/helper.rs [hunk 2]
```diff
@@ -0,0 +1,43 @@
...
+fn log2(mut n: u64) -> u64 {
+	let mut k = 0;
+	while n > 1 {
+		k += 1;
+		n >>= 1;
+	}
+	k
+}
+
+pub fn leaf_index_to_pos(index: u64) -> u64 {
```
HackFisher, line 11: I've made a PR to ckb_merkle_mountain_range:

https://github.com/nervosnetwork/merkle-mountain-range/pull/9/files

Should use this merged version for performance concern and code reuse.

frame/bridge/ethereum/relay/src/mmr.rs
```diff
@@ -0,0 +1,32 @@
...
+	fn merge(lhs: &Self::Item, rhs: &Self::Item) -> Self::Item {
+		let mut data = vec![];
+		data.append(&mut lhs.to_vec());
+		data.append(&mut rhs.to_vec());
+		hash(&data.as_slice())
+	}
+}
+
+/// Get mmr_size for the MMR model
+pub fn block_num_to_mmr_size(b: u64) -> u64 {
```
HackFisher, line 30: https://github.com/nervosnetwork/merkle-mountain-range/pull/9/files

primitives/ethereum-primitives/src/merkle.rs
```diff
@@ -0,0 +1,72 @@
+use array_bytes::{array_unchecked, fixed_hex_bytes_unchecked};
+use codec::{Decode, Encode};
+pub use ethereum_types::{H128, H512};
+use sp_io::hashing::sha2_256;
+use sp_runtime::RuntimeDebug;
+use sp_std::vec::Vec;
+
+#[derive(Clone, PartialEq, Eq, Encode, Decode, RuntimeDebug)]
+pub struct DoubleNodeWithMerkleProof {
```
HackFisher, line 9: This file is related to ethash proof, it would be better to rename the filename to indicated "ethash"

frame/bridge/ethereum/relay/src/helper.rs
yanganto: Thanks for remind.

frame/bridge/ethereum/relay/src/mmr.rs
yanganto: Thanks for remind.

frame/bridge/ethereum/relay/src/helper.rs
```diff
@@ -0,0 +1,43 @@
+// This file is part of merkle-mountain-range from nervosnetwork
+fn log2(mut n: u64) -> u64 {
```
HackFisher, line 2: @yanganto this file helper.rs can be deleted now

yanganto:
Thanks for review.

PR #170, opened 2020-07-01
Relayer Game Tests

AurevoirXavier:


frame/bridge/relayer-game/src/lib.rs
```diff
@@ -387,21 +414,24 @@ decl_module! {
...
-				<Proposals<T, I>>::take(game_id);
 				<Samples<T, I>>::take(game_id);
+				<LastConfirmeds<T, I>>::take(game_id);
+				<Headers<T, I>>::remove_prefix(game_id);
+				<Proposals<T, I>>::take(game_id);
 			};
 			let update_samples = |game_id, chain_len| {
-				<Samples<T, I>>::mutate(game_id, |samples| {
+				<Samples<T, I>>::mutate(game_id, |samples|
 					T::RelayerGameAdjustor::update_samples(
```
HackFisher, line 424: T::RelayerGameAdjustor::update_samples should emit a event telling that which kinds of mmr_proofs should be attached with header_thing for the next relayer submit& relay.

Maybe should be implemented in T::RelayerGameAdjustor

Could be some kind of

```
Event(Vec<(members, last_leaf)>)
```

PR #179, opened 2020-07-08
Fix mmr size in gen proof, Fixed #177

### LINKED ISSUE #177: mmr_size is different with position of last inserted block_number
https://github.com/darwinia-network/darwinia-common/blob/43673797183e75e3aa3c2bf651a09ee4e395e1f1/frame/header-mmr/src/lib.rs#L183

HackFisher:
Closes #177

### DIFF
--- a/frame/header-mmr/src/lib.rs
+++ b/frame/header-mmr/src/lib.rs
@@ -55,7 +55,7 @@ use codec::{Decode, Encode};
 use frame_support::{debug::error, decl_module, decl_storage};
 use sp_runtime::{
 	generic::{DigestItem, OpaqueDigestItemId},
-	traits::{Hash, Header},
+	traits::{Hash, Header, UniqueSaturatedInto},
 	RuntimeDebug,
 };
 use sp_std::{marker::PhantomData, prelude::*};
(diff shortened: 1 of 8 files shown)


AurevoirXavier:
```sh
16:12:55 in ~
λ curl http://localhost:9933 -H "Content-Type:application/json;charset=utf-8" -d '{ "jsonrpc": "2.0", "id": 1, "method": "headerMMR_genProof", "params": [0, 3] }'
{"jsonrpc":"2.0","result":{"mmrSize":"7","proof":"[0x313d79f4044b72f434aa2bf600a109c1d42e7b0717bf06f1f7d911eecb1c3e4c, 0xd3a82080ddb4a384a10060e34f3cf02f6094f9ff346f8edf8430a3ed8bda2021]"},"id":1}
```

frame/header-mmr/src/lib.rs
```diff
@@ -172,19 +179,18 @@ impl<T: Trait> MMRStore<T::Hash> for ModuleMMRStore<T> {
...
-			if block_number < mmr_block_number {
-				let pos = Self::position_of(block_number);
-				let mmr_header_pos = Self::position_of(mmr_block_number);
-
+			if block_number_of_member_leaf < block_number_of_last_leaf {
 				let store = <ModuleMMRStore<T>>::default();
-				let mmr = <MMR<_, MMRMerge<T>, _>>::new(mmr_header_pos, store);
+				let mmr_size = block_number_to_mmr_size(block_number_of_last_leaf);
+				let mmr = <MMR<_, MMRMerge<T>, _>>::new(mmr_size, store);
+				let pos = block_number_to_pos(block_number_of_member_leaf);
```
HackFisher, line 189: ```block_number_to_pos``` aka ```leaf_index_to_pos ``` from ckb mmr lib is not optimized, it can actually calculated by 2N - k - H - 1 or to_mmr_size(block) - H - 1, H is suffix zero counts(from last 1, exclusive)

ckb's version is ok to go for now.

clearloop: I've tried this algorithm yesterday morning:

```rust
pub fn pos(block_count: u64) -> u64 {
    // if suffix zero counts from last 1
    Self::mmr_size(block_count) - (((block_count + 1) >> 1).trailing_zeros() as u64) + 1
    // if suffix zero counts from last 0
    // Self::mmr_size(block_count) - ((block_count + 1).trailing_zeros() as u64) + 1
}
```

I'm not sure if my implementation is wrong, but it didn't work.

clearloop: The `Self::mmr_size` in that implementation is as same as `block_number_to_mmr_size` here, this algorithm works fine, tested with index (0..10)

frame/header-mmr/src/lib.rs
```diff
@@ -172,19 +179,18 @@ impl<T: Trait> MMRStore<T::Hash> for ModuleMMRStore<T> {
 impl<T: Trait> Module<T> {
 	impl_rpc! {
 		pub fn gen_proof_rpc(
-			block_number: T::BlockNumber,
-			mmr_block_number: T::BlockNumber,
+			block_number_of_member_leaf: u64,
+			block_number_of_last_leaf: u64,
 		) -> RuntimeDispatchInfo<T::Hash> {
```
clearloop, line 184: What if using short names of `block_number_of_member_leaf` and `block_number_of_last_leaf`, and complete the concepts in docs?

```rust
/// * member: block_number_of_member_leaf
/// * last_leaf: block_number_of_last_leaf
pub fn gen_proof_rpc(member: u64, last_leaf: u64) {}
```

The long arguments are easy to understand but hard to read.

frame/header-mmr/src/lib.rs
HackFisher: > I've tried this algorithm yesterday morning:
> 
> ```rust
> pub fn pos(block_count: u64) -> u64 {
>     // if suffix zero counts from last 1
>     Self::mmr_size(block_count) - (((block_count + 1) >> 1).trailing_zeros() as u64) + 1
>     // if suffix zero counts from last 0
>     // Self::mmr_size(block_count) - ((block_count + 1).trailing_zeros() as u64) + 1
> }
> ```
> 
> I'm not sure if my implementation is wrong, but it didn't work.

https://github.com/darwinia-network/darwinia-common/blob/aefdad9944e4a5c7611db3ba5892aa373ad1033c/frame/header-mmr/src/tests.rs#L148

PR #181, opened 2020-07-09
Hotfix Underflow

AurevoirXavier:
Due to the master branch are develop relayer game.

This hotfix should be merged into darwinia-master directly.


frame/staking/src/lib.rs
```diff
@@ -741,6 +741,7 @@ where
 						}
 					});
 				}
+				*active_deposit_ring -= slashable_active_ring;
```
HackFisher, line 744: Should be original(un-muted) ```slashable_deposit_ring ``` instead of ```slashable_active_ring ```?

PR #200, opened 2020-07-23
Substrate RC4

AurevoirXavier:


bin/node-template/runtime/src/lib.rs
```diff
@@ -100,58 +105,6 @@ pub mod impls {
 		}
 	}
 
-	/// Update the given multiplier based on the following formula
```
HackFisher, line 103: Did substrate have multiplier migration, changing from -1 to 0?

AurevoirXavier: It's OK with Crab.

https://github.com/paritytech/substrate/pull/6334/files#diff-c624291b4dc7bad33847a05c04bdb83eR255

HackFisher:
https://github.com/paritytech/substrate/pull/6334/files#diff-c624291b4dc7bad33847a05c04bdb83eR234

I worry that this migration codes will be removed in next releases(e.g. rc5), so need to keep an eye on this.

AurevoirXavier:
> https://github.com/paritytech/substrate/pull/6334/files#diff-c624291b4dc7bad33847a05c04bdb83eR234
> 
> I worry that this migration codes will be removed in next releases(e.g. rc5), so need to keep an eye on this.

https://github.com/darwinia-network/darwinia/issues/478

PR #230, opened 2020-08-13
Refactor relay and relayer-game modules

wuminzhe:
#188

PR #224, opened 2020-08-12
Upgrade Claims

AurevoirXavier:

PR #232, opened 2020-08-13
Common Library v1.0.0-alpha.2

AurevoirXavier:


frame/staking/src/lib.rs
```diff
@@ -341,60 +341,61 @@ mod migration {
...
+					.iter()
+					.fold(0.into(), |total_deposit, item| total_deposit + item.value)
+			{
+				if deposit_items.is_empty() {
+					// can not be, but better safe
+					*active_deposit_ring = 0.into();
 				} else {
-					// fix invalid data
-					*active_deposit_ring = Zero::zero();
+					*deposit_items = vec![TimeDepositItem {
```
HackFisher, line 394: Why not update active_deposit_ring instead?

Seems to be better than merging deposit items.

AurevoirXavier: Avoid `items.sum() > active`. But seems the data is ok.

AurevoirXavier:
Upgrade claims as suggestion: https://github.com/darwinia-network/darwinia-common/pull/224#issuecomment-674882244

AurevoirXavier:
Merge and upgrade #230

1. Optimize DB operations
2. Format `*.toml`
3. Format `*.rs`
4. Rename **API** `relay_headers` to `submit_proposal`
5. Rename **trait** `RelayerGame` to `RelayerGameProtocol`

bin/node-template/runtime/src/lib.rs
```diff
@@ -547,8 +559,8 @@ impl pallet_authorship::Trait for Runtime {
 parameter_types! {
 	pub const SessionsPerEra: SessionIndex = SESSIONS_PER_ERA;
 	pub const BondingDurationInEra: EraIndex = 14 * DAYS
-		/ (SESSIONS_PER_ERA as BlockNumber * BLOCKS_PER_SESSION);
-	pub const BondingDurationInBlockNumber: BlockNumber = 14 * DAYS;
+		/ (SESSIONS_PER_ERA as BlockNumber * BLOCKS_PER_SESSION) + 1;
+	pub const BondingDurationInBlockNumber: BlockNumber = 14 * DAYS + SESSIONS_PER_ERA as BlockNumber * BLOCKS_PER_SESSION;
```
HackFisher, line 563: The reasons for this change?

frame/balances/Cargo.toml
```diff
@@ -1,42 +1,50 @@
...
 default = ["std"]
+
 std = [
-	# crates
+	"crates-std",
+	"darwinia-std",
+	"substrate-std",
+]
+
+crates-std    = [
```
HackFisher, line 39: Old style looks simple and good, any reason for splitting them?

bin/node-template/runtime/src/lib.rs
AurevoirXavier: https://github.com/darwinia-network/darwinia/issues/495

Issue #188, opened 2020-07-15
Refactor suggestions regarding relay modules


HackFisher:
1. EthereumRelay implements EthereumRelayabe
2. Dispatch calls are all put into EthereumRelay, relay_header(EthereumHeaderThing). The parameter EthereumHeaderThing uses EthereumRelayabe codec into vec<v8> and is passed to RelayerGame’s submit_proposals. When using it, adjust the EthereumRelayabe interface. In this way, both apps and explorer can easily analyze the EthereumHeaderThing Type
3. As a more general and closed module, RelayerGame can manage its own storage, events and on_initialize/on_finalize, which are related to verifying the game sub-protocol.
4. EthereumRelay Dispatch Call calls RelayerGame to complete tasks related to verifying game sub-protocols.

HackFisher:
ConfirmBlockManagementError should be a Error, not a event.

AurevoirXavier:
Currently, the implementation is very ugly, very Inefficiency. 

Relay encode the header pass to Game, then Game call Relay to decode it. Such as `verify_raw_header_thing` is unnecessary. 

We should eventually separate the APIs. Refactor the Game's protocol

clearloop:
It might be a solution to implement a struct named `Relayable`:

```rust
struct Relayable {}

struct EthereumHeaderThing {}
struct TronHeaderThing{}

impl Into<Relayable> for EthereumHeaderThing{}
impl Into<Relayable> for TronHeaderThing{}
```

Or use a trait and proc_macro

```rust
struct Relayable {
  header: String,
  mmr: H256,
}

trait RelayableTrait {
   fn relayable() -> Relayable
}

/// -snippet-
/// a proc macro

#[derive(Relayable)]
struct EthHeaderThing{}

#[derive(Relayable)]
struct TronHeaderThing{}

```

HackFisher:
> Currently, the implementation is ugly, very inefficiency.
> 
> Relay encode the header things pass to Game, then Game call Relay to decode them. Such as `verify_raw_header_thing`, `verify_raw_header_things` are unnecessary under this architecture.
> 
> We should eventually separate the APIs. Refactor the Game's protocol.
> 
> Is possible to make Game an actual protocol? Like phragmen which defined in primitives.

Can we remove ```RawHeaderThing``` and make ```HeaderThing``` an associate type for both Relay and RelayerGame?

[...]

PR #242, opened 2020-08-18
Supports Submitting Raw Header Thing in Ethereum Relay API

clearloop:
## Desc

https://github.com/darwinia-network/darwinia-common/blob/d64934b659c140711c6508d8ec7ad9e37630f62a/frame/bridge/ethereum/relay/src/lib.rs#L187-L196

The refactor of `submit_proposal` in #232 breaks the API of `submit_proposal`, it would be nice to have a raw proposal option, otherwise, shadow and dj both need to refactor their `submit_proposal` APIs

---

Realized I kept the `EthHeaderThing` API during the last upgrading of shadow, need to confirm the workload of using the new `submit_proposal` API

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -502,9 +502,9 @@ dependencies = [
 
 [[package]]
 name = "blocking"
-version = "0.5.1"
+version = "0.5.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "2ad97f54e1c9030ad57fae2f5e3791447f48bb225ab20f9d80515e7017d04378"
+checksum = "ea5800d29218fea137b0880387e5948694a23c93fcdde157006966693a865c7c"
 dependencies = [
  "async-channel",
  "atomic-waker",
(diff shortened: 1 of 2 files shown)


AurevoirXavier:
Why the lock changed?

clearloop:
> Why the lock changed?

This API use `Vec<EthHeaderThing>` which has a different type definition in javascript comparing to `Vec<RawHeaderThing>` in the API in `RelayerGame`

| Rust | Javascript |
|-|-|
|EthHeaderThing| object |
|RawHeaderThing| string |

AurevoirXavier:
I mean the `Cargo.lock`

clearloop:
> I mean the `Cargo.lock`

...not sure, but it's okay to stash the `Cargo.lock` if we apply the `submit_raw_proposal` API

AurevoirXavier:
A workaround here. Due to #188

HackFisher (changes requested):
@AurevoirXavier Could we remove RawHeaderThing by using associate type, refer https://github.com/darwinia-network/darwinia-common/issues/188#issuecomment-676528053

PR #299, opened 2020-09-21
KTON Reward Tests

AurevoirXavier:


HackFisher (approved):

frame/staking/src/inflation_tests.rs
```diff
@@ -141,7 +141,31 @@ fn compute_total_payout_should_work() {
...
+		let kton_reward = (kton_reward * PRECISION).floor() / PRECISION;
+
+		// eprintln!("{:?}", kton_reward);
+
+		assert_eq!(kton_reward, *exp_kton_reward);
+	}
+}
+
+#[ignore]
+#[test]
```
AsceticBear, line 167: Ignore this test case?

AurevoirXavier: Just a print. No assertion inside.

PR #309, opened 2020-09-24
Refactor Dust Collector

AurevoirXavier:


AurevoirXavier:
Need to check the new account data is compatiable with the old one.

AurevoirXavier:
The data is ok.

frame/support/src/macros.rs
```diff
@@ -24,38 +24,38 @@ macro_rules! impl_account_data {
 			$($gtypebound)*
 		)?
 		{
-			pub free: Balance,
-			pub reserved: Balance,
+			pub free_ring: Balance,
```
HackFisher, line 27: Prefer to keep as it was.

Old name is ok, there could be migration works and apps compatible issues

HackFisher (changes requested):

AsceticBear (approved):

HackFisher (approved):

PR #334, opened 2020-10-16
Integrate frontier

AsceticBear:
## Desc

This is the first PR to integrate the frontier into the Darwinia network. 

## Finished Tasks

1. Clone the whole frontier repo and ensure it works in a template single node. The only change i made in the frontier-cloned now was the Trait for `ethereum-pallet`.

```sh
/// Trait for Ethereum pallet.
/// Old version
pub trait Trait: frame_system::Trait<Hash=H256> + pallet_balances::Trait + pallet_timestamp::Trait + pallet_evm::Trait {

New version for darwinia network
pub trait Trait: frame_system::Trait<Hash=H256> + darwinia_balances::Trait::<RingInstance> + pallet_timestamp::Trait + pallet_evm::Trait {
```

2. Change the default address mapping functions from original `HashedAddressMapping<BlakeTwo256>` to `ConcatAddressMapping`.

3. A new EVM precompiled contract `TransferBack`  added. The contract name and implementation might be changed later in future prs.


AurevoirXavier:
I think consensus + frame + rpc is enough.

AsceticBear:
> I think consensus + frame + rpc is enough.

OK, I am going to clear the useless files and commit later.

AurevoirXavier:
Which frontier version are we using?

I'll review this later.

AsceticBear:
> Which frontier version are we using?
> 
> I'll review this later.

Please check the first commit msg `Squashed 'frontier/' content from commit 51bd10f`.

HackFisher:
Better to move ```frontier/frame/ethereum``` to ```frame/evm```

For ```frontier/rpc``` and ```frontier/consensus```, maybe "rpc/evm" and "consensus/evm" is better.

AsceticBear:
> Better to move `frontier/frame/ethereum` to `frame/evm`
> 
> For `frontier/rpc` and `frontier/consensus`, maybe `rpc/evm` and `consensus/evm` is better.

Yeap, It looks more clear to put `frontier-related` pallets under `root/frame/`. 

How about putting `frontier/rpc`,  `frontier/frame/ethereum`, `frontier/consensus` into a single subdirectory of `root/frame` named vm. Because of these three pallets all tight-related with one topic.

```sh
darwinia-common
    - bin
    - frame
          - balances
          - evm
                - ethereum
                - consensus
                - rpc
```

HackFisher:
> > Better to move `frontier/frame/ethereum` to `frame/evm`
> > For `frontier/rpc` and `frontier/consensus`, maybe `rpc/evm` and `consensus/evm` is better.
> 
> Yeap, It looks more clear to put `frontier-related` pallets under `root/frame/`.
> 
> How about putting `frontier/rpc`, `frontier/frame/ethereum`, `frontier/consensus` into a single subdirectory of `root/frame` named vm. Because of these three pallets all tight-related with one topic.
> 
> ```shell
> darwinia-common
>     - bin
>     - frame
>           - balances
>           - dvm
>                 - ethereum
>                 - consensus
>                 - rpc
> ```

Yes, I think it is a good idea to do it in this way, just like another example in substrate:

https://github.com/paritytech/substrate/blob/master/frame/contracts/rpc/src/lib.rs

There are some rpc package from ethereum web3 which might be not directly to evm pallet, and more like primitives, can also moved to primitives folder

AsceticBear:
> > > Better to move `frontier/frame/ethereum` to `frame/evm`
> > > For `frontier/rpc` and `frontier/consensus`, maybe `rpc/evm` and `consensus/evm` is better.
> > 
> > 
> > Yeap, It looks more clear to put `frontier-related` pallets under `root/frame/`.
> > How about putting `frontier/rpc`, `frontier/frame/ethereum`, `frontier/consensus` into a single subdirectory of `root/frame` named vm. Because of these three pallets all tight-related with one topic.
> > ```shell
> > darwinia-common
> >     - bin
> >     - frame
> >           - balances
> >           - dvm
> >                 - ethereum
> >                 - consensus
> >                 - rpc
> > ```
> 
> Yes, I think it is a good idea to do it in this way, just like another example in substrate:
> 
> https://github.com/paritytech/substrate/blob/master/frame/contracts/rpc/src/lib.rs
> 
> There are some rpc package from ethereum web3 which might be not directly to evm pallet, and more like primitives, can also moved to primitives folder

So, I guess the final skeleton looks like the file tree I painted above just now?  If I am not misunderstand, I will work on this change. Thank you for making sure it again.

Cargo.lock
```diff
@@ -4444,6 +4740,35 @@ dependencies = [
  "sp-std 2.0.0",
 ]
 
+[[package]]
+name = "pallet-ethereum"
```
HackFisher, line 4744: 这个pallet是哪里用到的，感觉有可能不需要？

bin/node-template/node/src/service.rs
```diff
@@ -216,44 +223,11 @@ where
...
-	let justification_stream = grandpa_link.justification_stream();
-	let shared_authority_set = grandpa_link.shared_authority_set().clone();
 	let shared_voter_state = GrandpaSharedVoterState::empty();
 	let finality_proof_provider =
 		GrandpaFinalityProofProvider::new_for_service(backend.clone(), client.clone());
 	let import_setup = (babe_import.clone(), grandpa_link, babe_link.clone());
 	let rpc_setup = (shared_voter_state.clone(), finality_proof_provider.clone());
-	let babe_config = babe_link.config().clone();
-	let shared_epoch_changes = babe_link.epoch_changes().clone();
-	let rpc_extensions_builder = {
```
HackFisher, line 228: @AurevoirXavier pls help review this part, is it ok to remove it?

bin/node-template/runtime/src/lib.rs
```diff
@@ -1015,6 +1025,162 @@ impl darwinia_tron_backing::Trait for Runtime {
...
+}
+
+pub struct ConcatAddressMapping;
+
+impl AddressMapping<AccountId> for ConcatAddressMapping {
+	fn into_account_id(address: H160) -> AccountId {
+		let suffix = &BlakeTwo256::hash(&address[..])[0..12];
+
+		let mut data = [0u8; 32];
+		data[0..20].copy_from_slice(&address[..]);
```
HackFisher, line 1050: @wuminzhe please document detail dvm mapping address specification, considering adding specific prefix to indentify dvm mapping address from normal subkey address.

bin/node-template/runtime/src/lib.rs
```diff
@@ -1015,6 +1025,162 @@ impl darwinia_tron_backing::Trait for Runtime {
...
+		if cost > target_gas {
+			return Err(ExitError::OutOfGas);
+		}
+	}
+
+	Ok(cost)
+}
+
+/// Precompile for withdrawing from evm address
+pub struct Withdraw;
```
HackFisher, line 1079: this withdraw precompiled contract might be required by other runtimes too, better to move to library part.

HackFisher (changes requested):

bin/node-template/runtime/src/lib.rs
HackFisher: Maybe a checksum to prevent collision?

HackFisher:
@wuminzhe IMO ```Withdraw``` is not a good name for the precompiled contract, withdraw seems to be  "transfer to my own address", but this precompiled contract it to "transfer to any substrate address"

frame/dvm/rpc/Cargo.toml
```diff
@@ -0,0 +1,35 @@
...
+sp-storage = { git = "https://github.com/darwinia-network/substrate.git", branch = "common-library" } 
+sp-blockchain = { git = "https://github.com/darwinia-network/substrate.git", branch = "common-library" } 
+sc-service = { git = "https://github.com/darwinia-network/substrate.git", branch = "common-library" }
+sc-client-api = { git = "https://github.com/darwinia-network/substrate.git", branch = "common-library" }
+sc-rpc = { git = "https://github.com/darwinia-network/substrate.git", branch = "common-library" }
+sc-network = { git = "https://github.com/darwinia-network/substrate.git", branch = "common-library" }
+ethereum = { version = "0.3", features = ["codec"] }
+codec = { package = "parity-scale-codec", version = "1.0.0" }
+rlp = "0.4"
+pallet-ethereum = "0.1"
```
AsceticBear, line 33: @HackFisher  这里有一处依赖，要改成 DVM 内部的 ethereum，稍后我提交一下修改。

Cargo.lock
AsceticBear: 可以去掉的，请查看我在下方的评论。

bin/node-template/node/src/service.rs
AsceticBear: Yes, these code pieces needed more attention. The `rpc_extensions_builder` need a parameter `network`.

bin/node-template/runtime/src/lib.rs
AsceticBear: I recommend moving this code piece to ethereum pallet under dvm.

Cargo.lock
AsceticBear: Done.

frame/dvm/rpc/Cargo.toml
AsceticBear: Done.

HackFisher:
darwinia_address

AsceticBear:
Is it necessary to set the accountId Prefix to 11 length bytes? For the length of the checksum as short as possible?

4 bytes length Prefix might more understandable.  Why not leave an 8 length checksum at the end and keep 4 bytes length prefix at the head.

HackFisher:
> Is it necessary to set the accountId Prefix to 11 length bytes? For the length of the checksum as short as possible?
> 
> 
> 
> 4 bytes length Prefix might more understandable.  Why not leave an 8 length checksum at the end and keep 4 bytes length prefix at the head. 
> 
> 
> 
> 
> 
> 

1 byte is enough for checksum purpose.

The others are just leaving there for future extension such as version etc. Is is not necessary, but make sense IMO.

wuminzhe:
@HackFisher What precompile name will be better? `Transfer`?

HackFisher:
> @HackFisher What precompile name will be better? `Transfer`?

I have no idea, but I will suggest```BalanceTransfer``` or ```NativeTransfer``` if no more choices available.

frame/dvm/ethereum/Cargo.toml
```diff
@@ -28,8 +28,6 @@ libsecp256k1 = { version = "0.3", default-features = false }
 dvm-consensus-primitives = { path = "../consensus/primitives", default-features = false }
 dvm-rpc-primitives = { path = "../rpc/primitives", default-features = false }
 darwinia-support                  = { default-features = false, path = "../../../frame/support" }
-
-[dev-dependencies]
```
HackFisher: [dev-dependencies] are only used for tests, can be std, but frame dependencies must be no_std, be careful with ```default-features = false``` when moving between them

HackFisher (changes requested):

frame/dvm/ethereum/Cargo.toml
AsceticBear: What you said is too right. It took us a long time to debug the problem it causes.

AsceticBear:
> > @HackFisher What precompile name will be better? `Transfer`?
> 
> I have no idea, but I will suggest`BalanceTransfer` or `NativeTransfer` if no more choices available.

The final name: `NativeTransfer`.

bin/node-template/runtime/src/lib.rs
AsceticBear: Done.

bin/node-template/runtime/src/lib.rs
AsceticBear: Done.

AsceticBear:
@AurevoirXavier @HackFisher @wuminzhe  This PR based on the latest code is ready to review by all guys.

frame/dvm/ethereum/src/lib.rs
```diff
@@ -0,0 +1,464 @@
...
+	/// The overarching event type.
+	type Event: From<Event> + Into<<Self as frame_system::Trait>::Event>;
+	/// Find author for Ethereum.
+	type FindAuthor: FindAuthor<H160>;
+	type AddressMapping: AddressMapping<Self::AccountId>;
+	type RingCurrency: Currency<Self::AccountId>;
+}
+
+decl_storage! {
+	trait Store for Module<T: Trait> as Ethereum {
```
HackFisher, line 77: Do we need a different module storage prefix?

HackFisher (approved):

frame/dvm/ethereum/src/lib.rs
AsceticBear: Maybe not enough reason to do these changes at present.

AurevoirXavier (approved):

PR #359, opened 2020-10-28
Add an Voted Event for Guard

AurevoirXavier:


frame/bridge/ethereum/relay/src/lib.rs
```diff
@@ -133,6 +133,8 @@ decl_event! {
 		VerifyReceipt(AccountId, EthereumReceipt, EthereumHeader),
 		/// A relay header parcel got pended. [block number]
 		Pended(EthereumBlockNumber),
+		/// A guard voted. [ethereum block number, aye]
+		GuardVoted(EthereumBlockNumber, bool)
```
HackFisher, line 137: Missing a ```,``` here

HackFisher (changes requested):

HackFisher (approved):

PR #369, opened 2020-11-16
New test net setting

AsceticBear:


AurevoirXavier (approved):

bin/node/cli/src/service.rs
```diff
@@ -595,6 +591,6 @@ pub fn pangolin_new_full(
 }
 
 /// Create a new DRML service for a light client.
-pub fn pangolin_new_light(config: Configuration) -> Result<TaskManager, ServiceError> {
```
AsceticBear: Here, we create a new light client for pangolin, `pangolin_new_light`  is more appropriate than `drml_new_light`?

AurevoirXavier: All the runtime's new client can be generated from this

PR #377, opened 2020-11-19
Ethereum Backing Lock Assets

AurevoirXavier:


HackFisher:
Users who lock need to pay cross-chain transfer fees:

Before the full version complete, we use simple version, pay 50RING/Lock_TX, and the 50 RING fee will be transfered to the fee management account(like module account).

Ideas for full design version:
EcdsaAuthories and future delegate ethereum tx verifiers can to claim reward from the fee account.

HackFisher (approved):

PR #384, opened 2020-11-30
Add dvm web3 test

AsceticBear:
This PR contains a bunch of test scripts about the DVM RPC interface. I attached these tests to CI/CD to help us find potential problems early when making changes.

The CI workflow now:

Run Test                             ---> cargo test
Run Run DVM RPC Test    ---> cargo build + dvm rpc tests


.github/workflows/ci.yml
```diff
@@ -28,3 +28,30 @@ jobs:
 
       - name: Overall test
         run: cargo test
+      
+  dvm-rpc-test:
+    if: "contains(toJSON(github.event.head_commit.message), '[skip dvm]')"
```
wi1dcard, line 33: Should this be something like `!contains(..., '[skip dvm]')`?

.gitignore
```diff
@@ -21,3 +21,6 @@ release/
 .cargo-ok
 Cargo.toml.bak
 expand.rs
+
+# Js
+**/node_modules/
```
wi1dcard, line 26: It could (not 100% sure) work without the leading `**/`

AsceticBear: Indeed, it works well without `**/`.

.github/workflows/ci.yml
AsceticBear: A bit weird here. The second action workflow will be triggered when `if: "!contains(toJSON(github.event.head_commit.message), '[skip dvm]')"`. But, without `!`, it works well.

AsceticBear: I have tried hundred times with the head_commit flag to skip the DVM test for specific situations. Just ignore this feature at present.

HackFisher (approved):

HackFisher:
"Build and Test Expected — Waiting for status to be reported"

@wi1dcard why github report this which prevent merging?

wi1dcard:
> "Build and Test Expected — Waiting for status to be reported"
> 
> @wi1dcard why github report this which prevent merging?

Fixed. We set protected branches and "required status checks to pass before merging" in the repo settings, after renaming the workflow or job name, we'd have to change corresponding settings.

image

PR #385, opened 2020-12-01
Update dvm

AsceticBear:


HackFisher:
There are some problems with the code structure:

/frame/dvm/primitives should move to /primitives

/frame/dvm/primitives should move to /client

/frame/dvm/script_tests should move to /client/script_tests ?

HackFisher (approved):

Issue #380, opened 2020-11-20
mmr root stored in/for Header should be called Parent MMR Root


HackFisher:
https://github.com/darwinia-network/darwinia-common/blob/b9756a67bbb8e89f9af326ef5f98e66acd087b2d/frame/header-mmr/src/lib.rs#L62

- [x] Rename const var name to `PARENT_MMR_ROOT_LOG_ID `, do not need to change value.
- [ ] Update subscan display.

PR #391, opened 2020-12-07
Change name of mmr root to parent mmr root, related #380

HackFisher:


AurevoirXavier:
Should we change `pub mmr_root: Hash`, to `pub parent_mmr_root: Hash`

```rust
pub struct MerkleMountainRangeRootLog<Hash> {
	/// Specific prefix to identify the mmr root log in the digest items with Other type.
	pub prefix: [u8; 4],
	/// The merkle mountain range root hash.
	pub mmr_root: Hash,
}
```

HackFisher:
> Should we change `pub mmr_root: Hash` to `pub parent_mmr_root: Hash`
> 
> ```rust
> pub struct MerkleMountainRangeRootLog<Hash> {
> 	/// Specific prefix to identify the mmr root log in the digest items with Other type.
> 	pub prefix: [u8; 4],
> 	/// The merkle mountain range root hash.
> 	pub mmr_root: Hash,
> }
> ```

MerkleMountainRangeRootLog is a common struct, and has no directly connection to business meanings, so I was going to keep it as it is.

Another concern I want to confirm is: Changing the field of the struct will not affect storage codec, right?

AurevoirXavier:
> > Should we change `pub mmr_root: Hash` to `pub parent_mmr_root: Hash`
> > ```rust
> > pub struct MerkleMountainRangeRootLog<Hash> {
> > 	/// Specific prefix to identify the mmr root log in the digest items with Other type.
> > 	pub prefix: [u8; 4],
> > 	/// The merkle mountain range root hash.
> > 	pub mmr_root: Hash,
> > }
> > ```
> 
> MerkleMountainRangeRootLog is a common struct, and has no directly connection to business meanings, so I was going to keep it as it is.
> 
> Another concern I want to confirm is: Changing the field of the struct will not affect storage codec, right?

Yes, it won't. 

But if we use this struct in front-end or somewhere, it'll show `mmr_root`, but actually, it stores the `parent_mmr_root`

HackFisher:
> ```rust
> mmr_root
> ```

I changed `mmr_root` in the struct to `parent_mmr_root`.

AurevoirXavier (approved):

PR #398, opened 2020-12-10
Bump Version

AurevoirXavier:


frame/bridge/ethereum/backing/src/lib.rs
```diff
@@ -220,15 +220,6 @@ decl_module! {
 			0
 		}
 
-		fn on_runtime_upgrade() -> frame_support::weights::Weight {
```
HackFisher, line 223: Remove this will make darwinia has no effect on runtime upgrade?

HackFisher (changes requested):

frame/bridge/ethereum/backing/src/lib.rs
AurevoirXavier: We've 3 runtimes pangolin, crab, darwinia. So I recommend to do a custom runtime upgrade on their runtime.

AurevoirXavier:
@AsceticBear DVM CI failed.

AsceticBear (approved):

HackFisher (approved):

AurevoirXavier:
Migration https://github.com/darwinia-network/darwinia/pull/588/commits/d1dc15125f51a751999b400f8d86aa673fb8a92a

PR #411, opened 2020-12-18
deposit: current block number

AurevoirXavier:


AurevoirXavier:
Request from bridger. @wuminzhe

PR #414, opened 2020-12-21
fix: always return some

AurevoirXavier:
```rust
// T::Hash impl Default
pub MMRNodeList get(fn mmr_node_list): map hasher(identity) u64 => T::Hash;

fn get_elem(&self, pos: u64) -> Result<Option<T::Hash>> {
    Ok(Some(<Module<T>>::mmr_node_list(pos)))
}
```

This will always return a valid hash. Get mmr root will never fail. Even the size is wrong.


AsceticBear:
This change looks good to me.  I am not sure whether this will cause storage compatible issues.

AurevoirXavier:
> This change looks good to me. I am not sure whether this will cause storage compatible issues.

Won't.

AsceticBear (approved):

PR #416, opened 2020-12-25
Sync Authorities Set

AurevoirXavier:


frame/bridge/ethereum/backing/src/lib.rs
```diff
@@ -165,10 +172,11 @@ decl_storage! {
 	trait Store for Module<T: Trait> as DarwiniaEthereumBacking {
 		pub VerifiedProof
 			get(fn verified_proof)
-			: map hasher(blake2_128_concat) EthereumTransactionIndex => Option<bool>;
+			: map hasher(blake2_128_concat) EthereumTransactionIndex => bool = false;
```
HackFisher, line 175: Looks good, just to double check again:

Removing Option will not affect storage.

I remember last time @AsceticBear also asked the same question.

AurevoirXavier: I'm pretty sure. It won't affect the storage.

HackFisher (approved):

AurevoirXavier:
Need some test data.

Maybe next PR, if @WoeOm are not ready.

frame/bridge/ethereum/backing/src/lib.rs
AsceticBear: It would be more convincing if more analysis about the storage influence provided or local test.

HackFisher (approved):

WoeOm:
@wuminzhe

PR #425, opened 2020-12-30
feat(relay authorities): reset authorities

AurevoirXavier:
This call is dangerous. Should only be used in testnet.

### DIFF
--- a/frame/bridge/relay-authorities/src/lib.rs
+++ b/frame/bridge/relay-authorities/src/lib.rs
@@ -53,7 +53,7 @@ use frame_support::{
 	StorageValue,
 };
 use frame_system::ensure_signed;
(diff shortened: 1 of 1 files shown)


frame/bridge/relay-authorities/src/lib.rs
```diff
@@ -526,6 +533,28 @@ decl_module! {
 				<AuthoritiesToSign<T, I>>::put((message, signatures));
 			}
 		}
+
+		#[weight = 10_000_000]
+		pub fn kill_authorities(origin) {
+			T::ResetOrigin::ensure_origin(origin)?;
```
HackFisher, line 539: `ResetOrigin ` need to be carefully set in runtime

HackFisher (approved):

PR #434, opened 2021-01-06
Fix Relay Authorities Bugs

AurevoirXavier:
1. wrong term in the event
1. insufficient fee while paying for syncing-bridger

### DIFF
--- a/frame/bridge/ethereum/backing/src/lib.rs
+++ b/frame/bridge/ethereum/backing/src/lib.rs
@@ -326,16 +326,19 @@ decl_module! {
 			let beneficiary = Self::parse_authorities_set_proof(&proof)?;
 			let fee_account = Self::fee_account_id();
 			let sync_reward = T::SyncReward::get().min(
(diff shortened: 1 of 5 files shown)


HackFisher (approved):

frame/bridge/ethereum/backing/src/lib.rs
```diff
@@ -324,13 +324,21 @@ decl_module! {
...
-			)?;
-			T::EcdsaAuthorities::finish_authorities_change();
+			if !sync_reward.is_zero() {
+				T::RingCurrency::transfer(
+					&fee_account,
+					&beneficiary,
+					sync_reward,
+					KeepAlive
+				)?;
+				T::EcdsaAuthorities::finish_authorities_change();
```
HackFisher, line 340: Should move outside the block.

HackFisher (changes requested):

HackFisher (approved):

PR #436, opened 2021-01-08
Feat Staking Rebond

AurevoirXavier:
A lot of people unbond their assets accidentally.

With this feature, they don't need to wait 14 days and bond again.

### DIFF
--- a/bin/node/runtime/pangolin/src/weights/darwinia_staking.rs
+++ b/bin/node/runtime/pangolin/src/weights/darwinia_staking.rs
@@ -102,6 +102,12 @@ impl<T: frame_system::Trait> darwinia_staking::WeightInfo for WeightInfo<T> {
 			.saturating_add(T::DbWeight::get().writes(3 as Weight))
 			.saturating_add(T::DbWeight::get().writes((3 as Weight).saturating_mul(n as Weight)))
 	}
+	fn rebond(l: u32) -> Weight {
+		(46_569_000 as Weight)
+			.saturating_add((104_000 as Weight).saturating_mul(l as Weight))
+			.saturating_add(T::DbWeight::get().reads(4 as Weight))
+			.saturating_add(T::DbWeight::get().writes(3 as Weight))
+	}
 	fn set_history_depth(e: u32) -> Weight {
 		(0 as Weight)
 			.saturating_add((36_641_000 as Weight).saturating_mul(e as Weight))
--- a/frame/staking/src/default_weights.rs
+++ b/frame/staking/src/default_weights.rs
@@ -96,6 +96,13 @@ impl crate::WeightInfo for () {
 			.saturating_add(DbWeight::get().reads((5 as Weight).saturating_mul(n as Weight)))
 			.saturating_add(DbWeight::get().writes((3 as Weight).saturating_mul(n as Weight)))
 	}
+	fn rebond(l: u32) -> Weight {
+		(71316000 as Weight)
+			.saturating_add((142000 as Weight).saturating_mul(l as Weight))
+			.saturating_add(DbWeight::get().reads(4 as Weight))
+			.saturating_add(DbWeight::get().writes(3 as Weight))
+	}
+
 	fn set_history_depth(e: u32) -> Weight {
 		(0 as Weight)
 			.saturating_add((51901000 as Weight).saturating_mul(e as Weight))
--- a/frame/staking/src/lib.rs
+++ b/frame/staking/src/lib.rs
@@ -2015,7 +2015,7 @@ decl_module! {
 
 			ensure!(
 				!ledger.ring_staking_lock.unbondings.is_empty()
-					&& !ledger.kton_staking_lock.unbondings.is_empty(),
+					|| !ledger.kton_staking_lock.unbondings.is_empty(),
 				<Error<T>>::NoUnlockChunk
 			);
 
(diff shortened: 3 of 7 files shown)


bin/node/runtime/pangolin/src/weights/darwinia_staking.rs
```diff
@@ -102,6 +102,12 @@ impl<T: frame_system::Trait> darwinia_staking::WeightInfo for WeightInfo<T> {
 			.saturating_add(T::DbWeight::get().writes(3 as Weight))
 			.saturating_add(T::DbWeight::get().writes((3 as Weight).saturating_mul(n as Weight)))
 	}
+	fn rebond(l: u32) -> Weight {
+		(46_569_000 as Weight)
```
AsceticBear, line 106: How is the weight calculated here.

frame/staking/src/default_weights.rs
```diff
@@ -96,6 +96,13 @@ impl crate::WeightInfo for () {
 			.saturating_add(DbWeight::get().reads((5 as Weight).saturating_mul(n as Weight)))
 			.saturating_add(DbWeight::get().writes((3 as Weight).saturating_mul(n as Weight)))
 	}
+	fn rebond(l: u32) -> Weight {
+		(71316000 as Weight)
```
AsceticBear, line 100: The same question here.

bin/node/runtime/pangolin/src/weights/darwinia_staking.rs
AurevoirXavier: //! THIS FILE WAS AUTO-GENERATED USING THE SUBSTRATE BENCHMARK CLI VERSION 2.0.0

frame/staking/src/default_weights.rs
AurevoirXavier: //! THIS FILE WAS AUTO-GENERATED USING THE SUBSTRATE BENCHMARK CLI VERSION 2.0.0

AsceticBear: I checked the weight values set in the substrate latest code weight and found an inconsistency.

https://github.com/paritytech/substrate/blob/master/frame/staking/src/weights.rs#L362
```rs
fn rebond(l: u32, ) -> Weight {
		(49_966_000 as Weight)
			.saturating_add((92_000 as Weight).saturating_mul(l as Weight))
			.saturating_add(RocksDbWeight::get().reads(4 as Weight))
			.saturating_add(RocksDbWeight::get().writes(3 as Weight))
	}
```
Maybe i missed something here?

AurevoirXavier: Every network, every runtime has its own weight. Please learn more about **weight**.

AurevoirXavier: https://substrate.dev/docs/en/knowledgebase/learn-substrate/weight

AsceticBear: `Every network, every runtime, every implementation has its unique own weight` I agree with this view.

According to what you said above, the staking weight `THIS FILE WAS AUTO-GENERATED USING THE SUBSTRATE BENCHMARK CLI VERSION 2.0.0-rc6`, and the darwinia common does not seem to have a benchmark for the staking module separately. We have made some changes to the staking module, so the weight settings are inconsistent, it would be better to add some benchmark in the future.

AurevoirXavier: Already WIP. If you got some time. Keep going https://github.com/darwinia-network/darwinia-common/tree/xavier-benchmark

frame/staking/src/lib.rs
```diff
@@ -2480,19 +2517,17 @@ impl<T: Trait> Module<T> {
 		} = ledger;
 
 		if *active_ring != ring_staking_lock.staking_amount {
+			let origin_active_ring = ring_staking_lock.staking_amount;
```
HackFisher, line 2520: So, the origin_active_ring is exactly `ring_staking_lock.staking_amount`?

frame/staking/src/lib.rs
```diff
@@ -2480,19 +2517,17 @@ impl<T: Trait> Module<T> {
...
+
 			ring_staking_lock.staking_amount = *active_ring;
```
HackFisher, line 2522: The lock in staking is tracking the `active_ring`, so it is also working as original active ring before being updated.

frame/staking/src/lib.rs
```diff
@@ -3980,6 +4010,51 @@ where
...
+				} else {
+					let diff = plan_to_rebond.saturating_sub(rebonded);
+
+					rebonded = rebonded.saturating_add(diff);
+					*bonded = bonded.saturating_add(diff);
+					*amount = amount.saturating_sub(diff);
+				}
+
+				if rebonded >= plan_to_rebond {
+					break;
```
HackFisher, line 4041: Should not happen.

frame/staking/src/lib.rs
```diff
@@ -1984,6 +1984,55 @@ decl_module! {
...
+			ledger.ring_staking_lock.update(now);
+			ledger.kton_staking_lock.update(now);
+
+			ensure!(
+				!ledger.ring_staking_lock.unbondings.is_empty()
+					|| !ledger.kton_staking_lock.unbondings.is_empty(),
+				<Error<T>>::NoUnlockChunk
+			);
+
+			ledger.rebond(plan_to_rebond_ring, plan_to_rebond_kton);
```
HackFisher, line 2022: Just to make sure this is as expected, can add a testcase for this:

The sum of tokens in lock.unbonding might not enough for the plan_to_rebond.

HackFisher: If the unbonding is not enough, should we use free_balance, or throw an error, or keep as it is now?

HackFisher (approved):

PR #443, opened 2021-01-12
Checking on Sync Authorities

AurevoirXavier:


frame/bridge/ethereum/backing/src/lib.rs
```diff
@@ -644,13 +649,14 @@ impl<T: Trait> Module<T> {
...
 		proof_record: &EthereumReceiptProofThing<T>,
-	) -> Result<AccountId<T>, DispatchError> {
+	) -> Result<(Vec<EthereumAddress>, AccountId<T>), DispatchError> {
 		let log = {
 			let verified_receipt = T::EthereumRelay::verify_receipt(proof_record)
 				.map_err(|_| <Error<T>>::ReceiptProofInv)?;
 			let eth_event = EthEvent {
 				name: "SetAuthoritiesEvent".into(),
 				inputs: vec![
+					// TODO: Should we verify the nonce?
```
HackFisher, line 659: Yes, better to compare nonce with Term IMO.

frame/staking/src/lib.rs
```diff
@@ -2019,10 +2019,25 @@ decl_module! {
...
 
+			let active_ring = ledger.active_ring;
+			let active_kton = ledger.active_kton;
+
 			ledger.rebond(plan_to_rebond_ring, plan_to_rebond_kton);
 
 			Self::update_ledger(&controller, &mut ledger);
 
+			let rebond_ring = active_ring.saturating_sub(ledger.active_ring);
+			let rebond_kton = active_kton.saturating_sub(ledger.active_kton);
```
HackFisher, line 2030:
```suggestion
			let rebond_ring = ledger.active_ring.saturating_sub(active_ring);
			let rebond_kton = ledger.active_kton.saturating_sub(active_kton);
```

I think should be like this way?

HackFisher (changes requested):

frame/staking/src/lib.rs
AurevoirXavier: Ah yes.

HackFisher (approved):

PR #455, opened 2021-01-19
Precompile contract for withdraw

AsceticBear:
Add a new pre-compiled contract to achieve withdrawal from EVM account to Darwinia account.


frame/evm/src/runner/stack.rs
```diff
@@ -141,6 +144,12 @@ impl<T: Trait> RunnerT<T> for Runner<T> {
 		nonce: Option<U256>,
 		config: &evm::Config,
 	) -> Result<CallInfo, Self::Error> {
+		// native contracts call
+		if target == H160::from_str(TRANSFER_BACK_ADDR).unwrap() {
+			return T::TransferBack::execute(source, input, nonce)
```
HackFisher, line 149: Should have gas limit constraint?

bin/node/runtime/pangolin/src/lib.rs
```diff
@@ -1094,6 +1094,7 @@ impl darwinia_evm::Trait for Runtime {
 		darwinia_evm_precompile_simple::Sha256,
 		darwinia_evm_precompile_simple::Ripemd160,
 		darwinia_evm_precompile_simple::Identity,
+		darwinia_evm_precompile_withdraw::WithDraw<Self>,
```
hujw77, line 1097: Native contract may start at address `0x400` for compatibling evm.

frame/evm/precompile/withdraw/src/lib.rs
```diff
@@ -0,0 +1,68 @@
...
+
+		// Here is a tricky area that needs to be clarified.
+		// Once we trigger this precompile contract through an external transaction, the caller balance
+		// is transferred to the precompile contract account during the execution of the transaction in
+		// the stack runner and the transfer result is flushed to the storage after the transaction is
+		// completed. So here we only increase the address balance of the destination account and do not
+		// perform a decrease operation on the caller account balance, because the decrease operation is performed inside the stack runner.
+		T::Currency::deposit_creating(&input.dest, context_value);
+
+		return Ok((ExitSucceed::Returned, vec![], 500));
```
hujw77, line 46: `500` gas cost may be too low.

frame/evm/precompile/withdraw/src/lib.rs
```diff
@@ -0,0 +1,68 @@
...
+}
+
+impl<T: Trait> Precompile for WithDraw<T> {
+	fn execute(
+		input: &[u8],
+		_target_gas: Option<usize>,
+		context: &Context,
+	) -> core::result::Result<(ExitSucceed, Vec<u8>, usize), ExitError> {
+		// Decode input data
+		let input = InputData::<T>::decode(&input).unwrap();
```
hujw77, line 30: Error handling is missing.

frame/evm/precompile/withdraw/src/lib.rs
```diff
@@ -0,0 +1,68 @@
...
+
+		let helper = U256::from(10)
+			.checked_pow(U256::from(9))
+			.unwrap_or(U256::MAX);
+		let (context_value, _) = context.apparent_value.div_mod(helper);
+		let context_value = context_value.low_u128().unique_saturated_into();
+
+		// Here is a tricky area that needs to be clarified.
```
hujw77, line 38: So the address of `Withdraw` contract could be ignored when to calculate total supply  as the same as `0x0`.

bin/node/runtime/pangolin/src/lib.rs
AsceticBear: What you said is indeed a problem, in order to maintain maximum compatibility with ethereum, a minimum of 0x1-0x20 address space should be set aside. But in the dvm implementation, the precompile contract address is determined by the order set in runtime.

1. https://github.com/darwinia-network/darwinia-common/blob/master/primitives/evm/src/precompile.rs#L64
2. https://github.com/darwinia-network/darwinia-common/blob/master/bin/node/runtime/pangolin/src/lib.rs#L1092

Perhaps we can add some mock precompile contracts in the runtime to occupy this address space.

frame/evm/precompile/withdraw/src/lib.rs
AsceticBear: Any suggestions?

bin/node/runtime/pangolin/src/lib.rs
HackFisher: How about set some precompile placeholder instances here, which could be just as empty now, can be replaced later using runtime upgrade

frame/evm/precompile/withdraw/src/lib.rs
AsceticBear: Yes, 0x5 will be a black hole account, similar to 0x0. This account needs to be skipped when calculating the total number of rings, otherwise it will result in a higher calculation than the actual value.

bin/node/runtime/pangolin/src/lib.rs
AsceticBear: I think this approach works and I will create a separate issue on this problem.

AurevoirXavier:
Is this ready for merge?

AsceticBear:
> Is this ready for merge?

No, I still have some changes to make.

AsceticBear:
Close this pr for now,  I'll raise another regarding `withdraw precompile` later after #471 merged.

PR #457, opened 2021-01-20
Release 1.3.0

AurevoirXavier:


Cargo.lock
```diff
@@ -167,7 +167,7 @@ checksum = "43759f68cc29a37659a6b7a48a49674e03a888b6d41f36f4204c36d0d09a51de"
 
 [[package]]
 name = "array-bytes"
-version = "1.2.3"
+version = "1.3.0"
```
HackFisher, line 170:
```
[[package]]	[[package]]
name = "array-bytes"	name = "array-bytes"
version = "0.3.0"	version = "0.3.0"
source = "registry+https://github.com/rust-lang/crates.io-index"	source = "registry+https://github.com/rust-lang/crates.io-index"
checksum = "43759f68cc29a37659a6b7a48a49674e03a888b6d41f36f4204c36d0d09a51de"
```

There was another package name `array-bytes`, can we remove that one?

AurevoirXavier: Yes, there's a plan to remove the `primitives/array-bytes`.

Do you expect that in this version?

HackFisher (approved):

PR #461, opened 2021-01-25
Fix Slash

AurevoirXavier:


frame/bridge/relay-authorities/src/lib.rs
```diff
@@ -759,19 +759,34 @@ where
...
+						account_id, stake, ..
+					} in authorities.iter_mut()
 					{
-						<RingCurrency<T, I>>::slash(&account_id, stake);
+						if let None = signatures
+							.iter()
+							.position(|(authority, _)| authority == account_id)
+						{
+							if !stake.is_zero() {
+								<RingCurrency<T, I>>::slash(account_id, *stake);
```
HackFisher, line 774: Need a slash event here.

frame/bridge/relay-authorities/src/lib.rs
```diff
@@ -793,9 +808,19 @@ where
...
 		} else {
-			if let Some(signatures) =
-				<MMRRootsToSign<T, I>>::take(now - <SubmitDuration<T, I>>::get())
-			{
+			let at = now - <SubmitDuration<T, I>>::get();
+
+			if let Some(signatures) = <MMRRootsToSign<T, I>>::take(at) {
+				let _ = <MMRRootsToSignKeys<T, I>>::try_mutate(|keys| {
+					if let Some(position) = keys.iter().position(|key| key == &at) {
+						keys.remove(position);
```
HackFisher, line 816: Why need to remove the keys here ?

AurevoirXavier: We can't trace this data. Once the `SubmitDuration` changed.

```rust
let at = now.saturating_sub(<SubmitDuration<T, I>>::get());
<MMRRootsToSign<T, I>>::take(at);
```

HackFisher (approved):

PR #471, opened 2021-02-04
New stack runner 

AsceticBear:
There are two types of transaction executors in dvm, stack runner and builtin runner.

Previously, withdraw precompile contracts had to use the builtin runner to better interact with substrate storage, but the community has largely focused on the stack runner, and the builtin runner is far from stable. But now according to https://github.com/paritytech/frontier/pull/284, a new stack runner design has been proposed that can combine the features of both, and we need to integrate it.

In this pr, because we are using a conservative version of substrate, I had to use the evm crate branch from forked, in which I skipped evm 0.21 and committed 0.22 separately, see https://github.com/darwinia-network/evm/commits/darwinia for details.


frame/evm/src/lib.rs
```diff
@@ -534,4 +534,32 @@ impl<T: Trait> Module<T> {
 			Self::remove_account(address);
 		}
 	}
+
+	/// Withdraw fee.
+	pub fn withdraw_fee(address: &H160, value: U256) {
```
HackFisher, line 539: Is `withdraw_fee ` some additional trait from evm pallet? Looks good to me, just curious why it need an extra fee specific mutate api

HackFisher (approved):

frame/evm/src/lib.rs
AsceticBear: Here is just a simple encapsulation of balance modification.

PR #474, opened 2021-02-18
Withdraw precompile

AsceticBear:
A pre-compiled withdrawal precompile contract (evm address -> darwinia addresss).

The address is  0x5 temporarily, and I will raise another pull quest to modify this address separately later.

Js test scripts demo:

```js
const expect = require("chai").expect;
const Web3 = require("web3");

const web3 = new Web3("http://localhost:9933");

const addressFrom = "0x6Be02d1d3665660d22FF9624b7BE0551ee1Ac91b";
// substrate: '5ELRpquT7C3mWtjeqFMYqgNbcNgWKSr3mYtVi1Uvtc2R7YEx';
const privKey = "99B3C12287537E38C90A9219D4CB074A89A16E9CDB20BF85728EBD97C343E342";

const addressTo = "0x0000000000000000000000000000000000000005";

describe("Test Withdraw", function () {
    it("Withdraw", async function() {
        // target address -> 723908ee9dc8e509d4b93251bd57f68c09bd9d04471c193fabd8f26c54284a4b(5EeUFyFjHsCJB8TaGXi1PkMgqkxMctcxw8hvfmNdCYGC76xj)
        const input = "723908ee9dc8e509d4b93251bd57f68c09bd9d04471c193fabd8f26c54284a4b";

        const createTransaction = await web3.eth.accounts.signTransaction(
		{
		        from: addressFrom,
			to: addressTo,
                        gas: "5000000000",
                        data: input,
                        value: web3.utils.toWei("30", "ether"),
		},
		privKey
        );

        const createReceipt = await web3.eth.sendSignedTransaction(
            createTransaction.rawTransaction
        );
    }).timeout(10000);
});
```

Solidity scripts demo:

```sol
// delegateCall
pragma solidity ^0.8.0;
contract Test {
  function withdraw() public payable { 
    bytes32[3] memory output;
    bytes32[1] memory input;
    input[0] = hex"723908ee9dc8e509d4b93251bd57f68c09bd9d04471c193fabd8f26c54284a4b";
    assembly {
      if iszero(delegatecall(not(0), 0x05, input, 0x20, output, 0x40)) {
        revert(0, 0)
      }
    }
    // revert("attack!");
  }
}

// call
pragma solidity ^0.8.0;
contract Test {
  function withdraw() public payable { 
    bytes32[3] memory output;
    bytes32[1] memory input;
    input[0] = hex"723908ee9dc8e509d4b93251bd57f68c09bd9d04471c193fabd8f26c54284a4b";
    assembly {
      if iszero(call(not(0), 0x05, 40000000000000000000, input, 0x20, output, 0x40)) {
        revert(0, 0)
      }
    }
    // revert("attack!");
  }
}
```


script_tests/tests/1_test-transfer.js
```diff
@@ -83,4 +83,57 @@ describe("Test Transfer Balance", function () {
...
+		const balanceFrom = web3.utils.fromWei(await web3.eth.getBalance(addressFrom), "ether");
+		const balanceTo = await web3.utils.fromWei(await web3.eth.getBalance(addressTo), "ether");
+
+		expect(balanceFrom).to.be.equal("63.45678899999999999");
+		expect(balanceTo).to.be.equal("60");
+	});
+
+	it("Withdraw value from sender", async function () {
+		// target address = "723908ee9dc8e509d4b93251bd57f68c09bd9d04471c193fabd8f26c54284a4b(5EeUFyFjHsCJB8TaGXi1PkMgqkxMctcxw8hvfmNdCYGC76xj)";
+		// value = 30
```
hujw77, line 117:
```
$ seth sig "withdraw(bytes32,uint256)"  
0x040cf020
```
If there is a function selector, the encoding is different.
```shell
$ seth --to-uint256 30
0x000000000000000000000000000000000000000000000000000000000000001e
$ seth --to-bytes32 1e
0x1e00000000000000000000000000000000000000000000000000000000000000
```

AsceticBear: If I understand correctly, do you mean that the input test data should not be hardcoded here and should be generated by the tool?

frame/evm/precompile/withdraw/src/lib.rs
```diff
@@ -0,0 +1,102 @@
...
+		context: &Context,
+	) -> core::result::Result<(ExitSucceed, Vec<u8>, usize), ExitError> {
+		// Decode input data
+		let input = InputData::<T>::decode(&input)?;
+
+		let helper = U256::from(10)
+			.checked_pow(U256::from(9))
+			.unwrap_or(U256::MAX);
+		let value = input.value.saturating_mul(helper);
+		let from_address = T::AddressMapping::into_account_id(context.caller);
```
HackFisher, line 54: Does EVM automatically check the origin sender, more specifically, we can summary as following:

1. When the `from_address` is a normal address with a private key, EVM module check the signature before becoming `context.caller`.
2. When the `from_address` is a contract, EVM will automatic check `context.caller` when internal call happens between contracts.

script_tests/tests/1_test-transfer.js
hujw77: I mean that If there is a function selector `f(bytes32,uint256)`,  `input` should be encoded by that way.

HackFisher: @hujw77 Didn't get it.

`withdraw(bytes32,uint256)` is a wrapper contract in the description, it is not used here.

hujw77: I mean that `value` is encoded by bytes32 not uint256,  It should be more clearly.

HackFisher: By "32 bits for the withdrawal amount", @AsceticBear what't the encoded type for it?

AsceticBear: U256 little-endian.

frame/evm/precompile/withdraw/src/lib.rs
```diff
@@ -0,0 +1,102 @@
...
+			let mut dest_bytes = [0u8; 32];
+			dest_bytes.copy_from_slice(&data[32..64]);
+
+			let mut value_bytes = [0u8; 32];
+			value_bytes.copy_from_slice(&data[64..96]);
+
+			return Ok(InputData {
+				dest: <T as frame_system::Trait>::AccountId::decode(&mut dest_bytes.as_ref())
+					.map_err(|_| ExitError::Other("Invalid destination address".into()))?,
+				value: U256::from_little_endian(&value_bytes),
```
AsceticBear, line 97: @HackFisher @hujw77  Value type decode process here.

frame/evm/precompile/withdraw/src/lib.rs
AsceticBear: Yes, the check you said above will be done before entering the pre-compiled contract.

frame/evm/precompile/withdraw/src/lib.rs
HackFisher: This decode rules need to be documented as part of the spec.

HackFisher (changes requested):
For the encoding format of input, we should use the same ABI encoding methods as normal smart contract calldata to keep unified.

https://web3js.readthedocs.io/en/v1.3.0/web3-eth-abi.html#id6

The data payload in transaction is selector + calldata, but for precompiled contract, I have no idea whether we need to add extra bytes to use as an selector place holder. may need some tests here.

https://medium.com/swlh/understanding-data-payloads-in-ethereum-transactions-354dbe995371

HackFisher:
https://medium.com/@rbkhmrcr/precompiles-solidity-e5d29bd428c4

e.g.
```
delegatecall(not(0), 0x05, input, 0x60, output, 0x40)

staticcall(gasLimit, to, inputOffset, inputSize, outputOffset, outputSize)
```

In this delegatecall:
1. 0x05 is precompile address
2. 0x60 is input size, that is 96, if we change it to 0x40 (64), then the input received inside evm will be 64.
3. 0x40 is output size, which seems to be mismatch with the type here: `bytes32[3] memory output;`

HackFisher (approved):

frame/evm/precompile/withdraw/src/lib.rs
```diff
@@ -0,0 +1,112 @@
...
+		let context_value = context_value.low_u128().unique_saturated_into();
+
+		let result = T::Currency::transfer(
+			&contract_address,
+			&input.dest,
+			context_value,
+			ExistenceRequirement::AllowDeath,
+		);
+
+		ensure!(
```
hujw77, line 73: If support `delegatecall`, there is no guarantee that balance of context.address is zero.

HackFisher: Good catch.

hujw77: And if `target` address and `context.address` is same.

frame/evm/precompile/withdraw/src/lib.rs
```diff
@@ -0,0 +1,112 @@
...
+			<T as Trait>::Currency::free_balance(&contract_address)
+				== 0u128.unique_saturated_into(),
+			ExitError::Other("The contract balance should be zero after transfer".into())
+		);
```
HackFisher, line 77: Require to be removed.

HackFisher (changes requested):

frame/evm/precompile/withdraw/src/lib.rs
AsceticBear: I see what you mean, it's really not a good idea to check the balance here for a regular contract account.  I will remove this `ensure!`

AsceticBear: Done

hujw77 (commented):
LGTM

HackFisher (approved):

PR #480, opened 2021-02-22
rearrange precompile address

### LINKED ISSUE #467: WithDraw precompile address should be reset
Native contract may start at address `0x400` for compatibling evm.

_Originally posted by @hujw77 in https://github.com/darwinia-network/darwinia-common/pull/455#discussion_r566008756_

AsceticBear:
close #467


frame/evm/precompile/src/lib.rs
```diff
@@ -0,0 +1,43 @@
...
+	darwinia_evm_precompile_empty::Empty,      // 0x000000000000000000000000000000000000000c
+	darwinia_evm_precompile_empty::Empty,      // 0x000000000000000000000000000000000000000d
+	darwinia_evm_precompile_empty::Empty,      // 0x000000000000000000000000000000000000000e
+	darwinia_evm_precompile_empty::Empty,      // 0x000000000000000000000000000000000000000f
+	darwinia_evm_precompile_empty::Empty,      // 0x0000000000000000000000000000000000000010
+	darwinia_evm_precompile_empty::Empty,      // 0x0000000000000000000000000000000000000011
+	darwinia_evm_precompile_empty::Empty,      // 0x0000000000000000000000000000000000000012
+	darwinia_evm_precompile_empty::Empty,      // 0x0000000000000000000000000000000000000013
+	darwinia_evm_precompile_empty::Empty,      // 0x0000000000000000000000000000000000000014
+	darwinia_evm_precompile_withdraw::WithDraw<Runtime>, // 0x0000000000000000000000000000000000000015
```
HackFisher, line 42: Is 0x14 enough for reserve, since ethereum might add more Precompile in future too?

HackFisher (approved):

frame/evm/precompile/src/lib.rs
AsceticBear: For now, it's enough. If ethereum adds more than 20 precompiles contracts in the future, we can dynamically adjust it via runtime upgrade.

HackFisher: In that case, user contracts will have compatibility issues.

Considering that ethereum merely add precompile, this looks good to me.

AsceticBear: If this happens, our custom pre-compiled contract addresses will need to be modified, and the contracts that use the pre-compiled contracts will need to be modified. Because we can't predict how many contracts will be added to Ether in the future? Or let me add 10 more pre-compiled contracts that are a bit more scalable.

HackFisher: I think it is enough, let's keep as it is.

PR #490, opened 2021-02-25
Add kton precompile for evm

### LINKED ISSUE #477: KTON Precompile prototype and draft
```
KTON Precomiple {

// kton ledger
address -> balance // directly using KTON Currency module/pallet


// dangerous and be careful with to and data, usually must be WKTON and deposit.
transfer_and_call(to/* WKTON */, value, data /* e.g. deposit(..) */) {
    // transfer value to `to` address from msg.sender

    var from = msg.sender;
    
    // to.data(from, value)
}

withdraw..// to substrate地址

}


tx: {
    to, // KTON Precompile
    value, // 0
    nonce, // 
    data // input: method_sig/*transfer_and_call(...)*/ + call_data /*(WKTON_address, kton_value, "deposit(address, value)")*/
}

// https://etherscan.io/address/0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2#code
WKTON {
    function deposit(address from, uint256 value) public {
        require(msg.sender == KTON_precompile);
        balanceOf[msg.sender] += msg.value;
        Deposit(msg.sender, msg.value);
    }
}

tx: {
    to, // WKTON
    value, 
    nonce, 
    data
}
```

AsceticBear:
Users can use this pre-compiled contract to transfer the KTON balance to the pre-deployed dvm Wkton contract.

Close #477 

Here is a [demo](https://gist.github.com/AsceticBear/3f5baf2b238377a703833c3a83dd23e7) to show the use of the kton precompile.


hujw77 (commented):
A 'withdraw' function may be reqiured.

frame/evm/precompile/contracts/kton/src/lib.rs
```diff
@@ -0,0 +1,189 @@
...
+/// Kton Precompile Contract
+///
+/// The contract address: 0000000000000000000000000000000000000016
+pub struct Kton<T: Trait> {
+	_maker: PhantomData<T>,
+}
+
+impl<T: Trait> Precompile for Kton<T> {
+	// The execute process of Kton precompile, the input data consists of four parts:
+	// 1. Wkton solidity contract deployed address(20 bytes)
```
HackFisher, line 47: Could be any contract address?

frame/evm/precompile/contracts/kton/src/lib.rs
```diff
@@ -0,0 +1,189 @@
...
+use sp_std::vec::Vec;
+
+use core::str::FromStr;
+use darwinia_evm::{AddressMapping, Runner, Trait};
+use darwinia_evm_primitives::Precompile;
+use ethabi::{Function, Param, ParamType, Token};
+use evm::{Context, ExitError, ExitSucceed};
+use sp_std::borrow::ToOwned;
+
+const DEPOSIT_FUNC: &[u8] = b"deposit(address,uint256)";
```
HackFisher, line 36: Should not be hardcoded, this limited the use cases of this precompile

frame/evm/precompile/contracts/kton/src/lib.rs
```diff
@@ -0,0 +1,189 @@
...
+///
+/// The contract address: 0000000000000000000000000000000000000016
+pub struct Kton<T: Trait> {
+	_maker: PhantomData<T>,
+}
+
+impl<T: Trait> Precompile for Kton<T> {
+	// The execute process of Kton precompile, the input data consists of four parts:
+	// 1. Wkton solidity contract deployed address(20 bytes)
+	// 2. The function selector(4 bytes)
```
HackFisher, line 48: Why not call data?

frame/evm/precompile/contracts/kton/src/lib.rs
```diff
@@ -0,0 +1,189 @@
...
+			.unwrap_or(U256::MAX);
+		let deposit_func = &sha3::Keccak256::digest(&DEPOSIT_FUNC)[0..4];
+		// decode input
+		let input = InputData::decode(input)?;
+
+		let con_caller = T::AddressMapping::into_account_id(context.caller);
+		if hex::encode(input.func_selector) == hex::encode(&deposit_func) {
+			// 1. Transfer kton from sender to kton erc20 contract
+			let wkton_account_id = T::AddressMapping::into_account_id(input.wkton_address);
+			let transfer_value = input.value.saturating_mul(helper);
```
HackFisher, line 67: What's the decimal of the `transfer_value`?

frame/evm/precompile/contracts/kton/src/lib.rs
```diff
@@ -0,0 +1,189 @@
...
+			.checked_pow(U256::from(9))
+			.unwrap_or(U256::MAX);
+		let deposit_func = &sha3::Keccak256::digest(&DEPOSIT_FUNC)[0..4];
+		// decode input
+		let input = InputData::decode(input)?;
+
+		let con_caller = T::AddressMapping::into_account_id(context.caller);
+		if hex::encode(input.func_selector) == hex::encode(&deposit_func) {
+			// 1. Transfer kton from sender to kton erc20 contract
+			let wkton_account_id = T::AddressMapping::into_account_id(input.wkton_address);
```
HackFisher, line 66: what's the difference between `input.address` and `wkton_account_id `?

HackFisher (changes requested):

HackFisher:
I would like to propose a new spec of the `transfer_and_call` input data.

1. The `to` address, corresponding subkey of  this targetevm address will receive kton value.
2. The `kton_value`, decimal is 9, the same as KTON currency.
3. `call_data`, any encoded call data, if `to` is a smart contract address, `T::Runner::call` will call with this call data.


`withdraw` input data might be quite similar to `Withdraw` Precompile.

AsceticBear:
> 3\. Is there a way to tell whether a dvm address is a smart contract address or not?

This can be done by checking if the address account contains a contract code.

frame/evm/precompile/contracts/kton/src/lib.rs
AsceticBear: This address refers to the wkton contract address deployed in advance on the dvm. If this address is not correct when executing a call in precompile, it will not succeed.

frame/evm/precompile/contracts/kton/src/lib.rs
HackFisher: Let's not use two addresses here, input.address can do redirection to transfer the kton value to any other address he would like.

In the case of WKTON, `wkton_account_id` is encoded as part of the `call_data`, it is `to`(WKTON here) to decide how to parse the `call_data` and how to use it. We do not need to specify `wkton_account_id` in precompile input.

If the target `to` address is not WKTON, there could be no such redirection address.

@hujw77 Would you please help provide a WKTON implementation for test case purpose?

AsceticBear: https://github.com/evolutionlandorg/evo-deploy/blob/main/src/wrapper/WKTON.sol

hujw77: https://github.com/evolutionlandorg/evo-deploy/blob/main/src/wrapper/WKTON.sol
@HackFisher Need to be reviewed too.

HackFisher: @hujw77 WKTON is better to move to darwinia repository, can also be shared as common used contract for other applications.

https://github.com/evolutionlandorg/evo-deploy/blob/f340fcf67219f77ee788e880099fd378920198dc/src/wrapper/WKTON.sol#L21

Need to check the decimal convert, the kton_value from KTON Precompile is 9 now.

Need to check whether need an extra `receiver` address in `function deposit(address from, uint256 value)`, with or without it, both ok for me. Now I prefer be simpler, aka. without `receiver`wkton_account_id, as it is.

frame/evm/precompile/contracts/kton/src/lib.rs
```diff
@@ -0,0 +1,189 @@
...
+			let precompile_address = H160::from_str(KTON_PRECOMPILE).unwrap();
+			T::Runner::call(
+				precompile_address,
+				input.wkton_address,
+				raw_input.to_vec(),
+				U256::zero(),
+				target_limit.unwrap_or_default() as u32,
+				None,
+				None,
+				T::config(),
```
HackFisher, line 87:
```suggestion
				precompile_address,   // msg.sender, KTON precompile
				to_address,
				call_data,
				U256::zero(), // RING, value
				target_limit.unwrap_or_default() as u32,
				None,
				None,
				T::config(),
```

frame/evm/precompile/contracts/kton/src/lib.rs
```diff
@@ -0,0 +1,189 @@
...
+			T::KtonCurrency::transfer(
+				&con_caller,
+				&wkton_account_id,
+				transfer_value.low_u128().unique_saturated_into(),
+				ExistenceRequirement::AllowDeath,
+			)
+			.map_err(|_| ExitError::Other("Transfer in Kton precompile failed".into()))?;
+
+			// 2. Call wkton sol contract deposit
+			let raw_input = make_deposit_input(input.address, input.value)?;
```
HackFisher, line 77: e.g. calldata is `selector("deposit(from, value)")`

```suggestion
			let raw_input = make_deposit_input(calldata, input.address, input.value)?; 
```

AsceticBear:
> A 'withdraw' function may be reqiured.

Done

HackFisher (approved):
LGTM

hujw77 (commented):
LGTM

PR #498, opened 2021-03-04
DVM Updates

AsceticBear:
NOTE：

This pr is now ready for review, but there is still a little bit of work left, i.e., upgrading the substrate to 3.0 and the codec version, which needs to wait for the frontier library to be updated, but you can review it first because there are more changes and I'd like to leave more time for review.


client/dvm/consensus/src/lib.rs
```diff
@@ -128,24 +132,44 @@ where
...
+
+			let mapping_commitment = dvm_db::MappingCommitment {
+				block_hash: hash,
+				ethereum_block_hash: post_hashes.block_hash,
+				ethereum_transaction_hashes: post_hashes.transaction_hashes,
+			};
+			let res = self.backend.mapping_db().write_hashes(mapping_commitment);
+			if res.is_err() {
+				trace!(target: "frontier-consensus", "{:?}", res);
+			}
```
HackFisher, line 153: There are changes related to client db here. Do we need to do client side db migration when we incorporate dvm into Crab Network?

AsceticBear: At this stage, the crab network does not have an integrated DVM solution. I don't think migration is needed if the upgraded database solution is chosen directly when integrating DVM to crab, right? But there is another problem, if the DVM already running on crab has data incompatibility in the future, the way of data migration may be different from the way of substrate migration, which needs further thought.

HackFisher: Let's do such a test on Pangolin.

1. Upgrade to simulate supporting new added DVM to Pangolin. (Clean DVM from Pangolin before that)
2. DVM already running on pangolin, do client backend migration if there are incompatibility datas or upgrades from Frontier.

AsceticBear: #512 here.

client/dvm/rpc/src/eth.rs
```diff
@@ -484,10 +484,7 @@ where
 				.get(&schema)
 				.unwrap_or(&self.fallback)
 				.storage_at(&id, address, index)
-				.ok_or(internal_err(
-					"Fetching account storage through override failed",
-				))?
-				.into());
+				.unwrap_or_default());
```
AsceticBear, line 487: It would be better to apply this change to the frontier.

client/dvm/consensus/Cargo.toml
```diff
@@ -4,20 +4,21 @@ description = "Consensus for darwinia dvm"
 edition     = "2018"
 homepage    = "https://darwinia.network/"
 license     = "GPL-3.0"
-name        = "dvm-consensus"
+name        = "dc_consensus"
```
HackFisher, line 7: Most crates are using slash `-`, it is weird to use underscore `_` here.

Cargo.lock
```diff
@@ -1410,7 +1410,7 @@ dependencies = [
 name = "darwinia-evm-precompile-blake2"
 version = "2.0.0"
 dependencies = [
- "darwinia-evm-primitives",
+ "dp-evm",
```
HackFisher, line 1413: @AurevoirXavier Are you good with this prefix convention?

bin/node/cli/src/rpc.rs
```diff
@@ -206,13 +219,30 @@ where
 	io.extend_with(BalancesApi::to_delegate(Balances::new(client.clone())));
 	io.extend_with(HeaderMMRApi::to_delegate(HeaderMMR::new(client.clone())));
 	io.extend_with(StakingApi::to_delegate(Staking::new(client.clone())));
+
+	let mut overrides = BTreeMap::new();
+	overrides.insert(
+		EthereumStorageSchema::V1,
```
HackFisher, line 225: Need documents, especially for future update, old Schema should not be deleted for history blocks. (as we summarized)

bin/node/cli/src/service.rs
```diff
@@ -161,6 +171,27 @@ fn set_prometheus_registry(config: &mut Configuration) -> Result<(), ServiceErro
 	Ok(())
 }
 
+pub fn open_frontier_backend(config: &Configuration) -> Result<Arc<dc_db::Backend<Block>>, String> {
+	let config_dir = config
+		.base_path
+		.as_ref()
+		.map(|base_path| base_path.config_dir(config.chain_spec.id()))
+		.unwrap_or_else(|| {
+			BasePath::from_project("", "", &crate::cli::Cli::executable_name())
```
HackFisher, line 180: Check and test data dir

client/dvm/rpc/src/eth.rs
HackFisher: Is frontier has a pending PR waiting for being merged?

Please add some document here with the link to related issue or PR.

frame/dvm/src/lib.rs
```diff
@@ -111,7 +128,10 @@ decl_storage! {
 	}
 	add_extra_genesis {
 		build(|_config: &GenesisConfig| {
-			<Module<T>>::store_block();
+			<Module<T>>::store_block(false);
+
+			// Initialize the storage schema at the well known key.
+			frame_support::storage::unhashed::put::<EthereumStorageSchema>(&PALLET_ETHEREUM_SCHEMA, &EthereumStorageSchema::V1);
```
HackFisher, line 134: If `V1` is setup in `on_genesis`, there could be issues with middle-flight running network(Crab&Darwinia), which means we might need extra migration for those existing networks to upgrade?

frame/dvm/src/lib.rs
```diff
@@ -220,11 +247,20 @@ decl_module! {
...
+			);
 		}
 
 		fn on_initialize(_block_number: T::BlockNumber) -> Weight {
 			Pending::kill();
+			if let Ok(log) = dp_consensus::find_pre_log(&frame_system::Module::<T>::digest()) {
+				let PreLog::Block(block) = log;
+
+				for transaction in block.transactions {
+					let _ = Call::<T>::transact(transaction).dispatch_bypass_filter(RawOrigin::None.into());
```
HackFisher, line 261: This looks like it will always fail due to the check in dispatch call, as we discussed.

HackFisher (changes requested):
@AurevoirXavier @wuminzhe A lot of changes, please help review this PR.

client/dvm/rpc/src/eth.rs
AsceticBear: Not yet until now. let me do a fix about this issue.

client/dvm/consensus/Cargo.toml
AurevoirXavier: Recommend to use `-`

Cargo.lock
AurevoirXavier: LGTM.

Need a discussion here for further usage. https://github.com/darwinia-network/primitives/issues/1

client/dvm/rpc/src/eth.rs
AsceticBear: https://github.com/paritytech/frontier/pull/316

bin/node/cli/src/service.rs
AsceticBear: Based on the current implementation:
```rs
let database_dir = config_dir.join("frontier").join("db");
```
the backend db path is `/data/alice/chains/pangolin/frontier/db`， maybe we should change it to `/data/alice/chains/pangolin/dvm/db`.  I need your advice at this point. @HackFisher

frame/dvm/src/lib.rs
AsceticBear: Yes, when integrating DVM to crab, the storage entry, EthereumStorageSchema, does not exist on other networks, so you need to add this entry manually, I think migration can do this, create a new issue?

HackFisher: I see, I created an issue here: https://github.com/darwinia-network/darwinia/issues/624

bin/node/cli/src/service.rs
HackFisher: Yes, let's change it to `config_dir.join("dvm").join("db");`

frame/dvm/src/lib.rs
AsceticBear: This question is not answered in the upstream library at the moment. Let me create an issue to track, and do not discuss it too much in this pr, it's already huge enough.

client/dvm/db/Cargo.toml
```diff
@@ -0,0 +1,21 @@
+[package]
+authors     = ["Darwinia Network <hello@darwinia.network>"]
+description = "Dvm database backend"
+edition     = "2018"
+homepage    = "https://darwinia.network/"
+license     = "GPL-3.0"
+name        = "dc-db"
+readme      = "README.md"
+repository  = "https://github.com/darwinia-network/darwinia-common/"
+version     = "1.3.0"
```
HackFisher, line 10: We've bumped all the crate version in common library to `2.0.0`.

Related crates in this PR should be bumped too.

HackFisher (changes requested):

HackFisher (approved):

AurevoirXavier:
Ready for review now?

AsceticBear:
> Ready for review now?

Go!

AurevoirXavier (approved):

PR #503, opened 2021-03-05
Bridge registration protocol(Issuing and MappingTokenFactory)

xiaoch05:


xiaoch05:
moved to https://github.com/darwinia-network/darwinia-common/pull/546 which based on the latest version v2.0.0

PR #532, opened 2021-03-14
Update Parity Common

AurevoirXavier:
Because mater is on codec 1.3.7.

I decide to make this change on rococo-v1 first.

Later the rococo-v1 will merge into master.

@HackFisher 

And due to the frontier issue, this node can't be compile. But you could run `cargo t` under the modules' folder.


primitives/ethereum-primitives/src/header.rs
```diff
@@ -174,7 +174,10 @@ impl EthereumHeader {
 					array_bytes::hex2array_unchecked!(parse_value_unchecked(s), 32).into();
 			}
 		}
-		eth_header.seal = vec![rlp::encode(&mix_hash), rlp::encode(&nonce)];
+		eth_header.seal = vec![
+			rlp::encode(&mix_hash).to_vec(),
```
HackFisher, line 178: Is there difference between `rlp::encode(&mix_hash)` and `rlp::encode(&mix_hash).to_vec()` ?

HackFisher:
@AurevoirXavier Would you add the link to the frontier issue which prevent compile?

primitives/ethereum-primitives/src/header.rs
AurevoirXavier: `rlp::encode` now returns a bytes thing, we've to convert it to `Vec` manually.

HackFisher:
@AurevoirXavier There are conflicting files, need to be resolved.

And, could we merge https://github.com/darwinia-network/darwinia-common/pull/536 first?

HackFisher:
Why copy  https://github.com/darwinia-network/ethash as part crate, not using it as dependency?

Both ok for me, but using it as dependency might be easier for upstream updates.

AurevoirXavier:
> Why copy https://github.com/darwinia-network/ethash and include it as sub crate, not using it as dependency?
> 
> Both ok for me, but using it as dependency might be easier for tracking upstream updates.

Just for testing. But I forgot to remote them.

HackFisher (approved):

PR #546, opened 2021-03-18
Bridge registration protocol(Issuing and MappingTokenFactory)#v2.0.0

xiaoch05:


frame/dvm/src/lib.rs
```diff
@@ -169,78 +175,7 @@ decl_module! {
...
-				used_gas,
-				logs_bloom: status.clone().logs_bloom,
-				logs: status.clone().logs,
-			};
-
-			Pending::append((transaction, status, receipt));
-
-			Self::deposit_event(Event::Executed(source, contract_address.unwrap_or_default(), transaction_hash, reason));
-			Ok(Some(T::GasWeightMapping::gas_to_weight(used_gas.unique_saturated_into())).into())
+			Self::raw_transact(source, transaction)
```
AsceticBear, line 178: The changes here may make synchronizing frontier more complicated in the future, see https://github.com/paritytech/frontier/pull/324. 

Why change to `raw_transact` here?

xiaoch05: raw_transact will not verify the signature of the transaction.
The system account can't sign the transaction but send it directly.

AsceticBear: Just take verify signature part out of `do_transact` can also work, right? No need to do more abstract `DvmRawTransactorT` in this approach.

frame/evm/src/lib.rs
```diff
@@ -254,6 +254,11 @@ impl GasWeightMapping for () {
 	}
 }
 
+/// handle from evm contract
```
AsceticBear, line 257: /// ~~handle~~ from evm contract

frame/evm/precompile/contracts/issuing/Cargo.toml
```diff
@@ -0,0 +1,57 @@
...
+readme      = "README.md"
+repository  = "https://github.com/darwinia-network/darwinia-common/"
+version     = "2.1.0"
+
+[dependencies]
+# crates
+codec                    = { default-features = false, package = "parity-scale-codec", version = "2.0.1" }
+dp-evm  = { default-features = false, path = "../../../../../primitives/evm" }
+evm                     = { version = "0.24.0", default-features = false, features = ["with-codec"] }
+ripemd160                = { default-features = false, version = "0.9.1" }
```
AsceticBear, line 17: Please remove these useless dependencies and keep `*.toml` clean.  such as `ripemd160`

frame/bridge/ethereum/issuing/src/lib.rs
```diff
@@ -0,0 +1,431 @@
...
+				.saturated_into(),
+		);
+		Ok(())
+	}
+}
+
+impl<T: Config> ContractHandler for Module<T> {
+	/// handle
+	fn handle(address: H160, caller: H160, input: &[u8]) -> DispatchResult {
+		if MappingFactoryAddress::get() == caller.0.into() {
```
AsceticBear, line 412: if `MappingFactoryAddress::get() != caller.0.into()`, return dispatch error would be better?

frame/bridge/ethereum/issuing/src/lib.rs
```diff
@@ -0,0 +1,431 @@
...
+	}
+}
+
+decl_event! {
+	pub enum Event<T>
+	where
+		AccountId = AccountId<T>,
+	{
+		/// test
+		Test(AccountId),
```
AsceticBear, line 126: `Test(AccountId` can be removed now?

frame/bridge/ethereum/issuing/contract/src/lib.rs
```diff
@@ -0,0 +1,379 @@
...
+	RawLog, Result as AbiResult,
+};
+use ethereum_types::{Address as EthereumAddress, H160, H256, U256};
+
+pub use ethabi::{Event, Log};
+use sp_std::vec::Vec;
+
+pub struct Abi;
+
+pub struct Address([u8; 20]);
```
AsceticBear, line 37: Change to `sp_core::H160` or `ethereum_type::H160` also works?

frame/dvm/src/lib.rs
xiaoch05: yes, can we modify `do_transact `? it may be used by others.

AsceticBear: I come up with another idea:

1. keep `do_transact` as usual, and add flag to indicate whether to verify signature.

frame/bridge/ethereum/issuing/contract/src/lib.rs
xiaoch05: I can see the `dvm` and other pallets have the same version of this type in the latest version.
We can try to change it to H160 directly now.

AsceticBear:
You could run `cargo format`, and `cargo clippy` in the crate you added and do some optimization follow Clippy tips, then commit them together.

frame/evm/precompile/contracts/issuing/Cargo.toml
AsceticBear: Please check not only this file but also other *.toml files.

frame/dvm/src/lib.rs
xiaoch05: you mean we add a flag as a parameter in `do_transact`? I think it's a good idea. it can keep the code structure as consistent as possible with the original.

AsceticBear: Yes, exactly.

Cargo.toml
```diff
@@ -14,6 +14,8 @@ members = [
 	"frame/bridge/crab/backing",
 	"frame/bridge/crab/issuing",
 	"frame/bridge/ethereum/backing",
+    "frame/bridge/ethereum/issuing",
+    "frame/bridge/ethereum/issuing/contract",
```
HackFisher, line 18: Format issue

frame/bridge/ethereum/issuing/src/lib.rs
```diff
@@ -0,0 +1,429 @@
...
+					<Error<T>>::ReceiptProofInv
+				})?;
+
+			let register_event = Abi::register_event();
+			let backing_event = Abi::backing_event();
+			let log_entry = verified_receipt
+				.logs
+				.into_iter()
+				.find(|x| {
+					x.address == backing &&
```
HackFisher, line 180: Anyone can call this dispatch call, and the backing is passed by the user, so he can specify any backing address as he want.

So it is useless for this validation, and backing can be removed from call parameters.

And as discussed, to prevent too many backing and mapping tokens spam in the bridge, we will only use one backing for each source chain.

xiaoch05: how can we change this parameter as network Id?  and then all other ethereum-like chains can be processed by a common way.

HackFisher: Yes, we need a network id in the event and proof. Since we will verify the proof and extract that value from the proof.

Passing the network id as call parameter is meaningless for the same reason.

HackFisher: I need to correct, if consider the network id, we may need that in the call parameter since we have to know which network's relay will be used to verify this receipt, only after that, we can tell the receipt format and info.

But for here in ethereum, we do not need network_id in the event/receiot.

frame/dvm/src/lib.rs
```diff
@@ -435,6 +441,28 @@ impl<T: Config> Module<T> {
 			used_gas.unique_saturated_into(),
 		))
 		.into())
+
+	}
+
+    pub fn raw_call(
```
AsceticBear, line 447: Why not use `do_transact` to do this `call`?

xiaoch05: `do_transact` will check nonce, check gas price, and save pending. I think it's not suitable here.

AsceticBear: There is another way can achieve this. https://github.com/darwinia-network/darwinia-common/blob/master/bin/node/runtime/pangolin/src/lib.rs#L798

The advantage of your implementation is that the external crate just only interacts directly with the dvm pallet, which I think is acceptable.

xiaoch05: yes, `do_call` seems better.

frame/evm/precompile/contracts/issuing/src/lib.rs
```diff
@@ -0,0 +1,47 @@
...
+
+/// Issuing Precompile Contract, used to burn mapped token and generate a event proof in darwinia
+///
+/// The contract address: 0000000000000000000000000000000000000017
+pub struct Issuing<T: Config> {
+	_maker: PhantomData<T>,
+}
+
+impl<T: Config> Precompile for Issuing<T> {
+	/// Input data: 32-bit substrate withdrawal public key
```
AsceticBear, line 37: This comment is not suitable for this precompile? 

```rs
fn handle(address: H160, caller: H160, input: &[u8]) -> DispatchResult {
	ensure!(MappingFactoryAddress::get() == caller, <Error<T>>::AssetAR);
	if input.len() == 3 * 32 {}
```

Please add more comments here about this input usage for code readers in the future.

frame/evm/src/lib.rs
```diff
@@ -285,6 +290,8 @@ pub trait Config: frame_system::Config + pallet_timestamp::Config {
 	type Runner: Runner<Self>;
 	/// The account basic mapping way
 	type AccountBasicMapping: AccountBasicMapping;
+	/// Contract caller
```
AsceticBear, line 293: `/// Contract Handler` is better?

frame/bridge/ethereum/issuing/Cargo.toml
```diff
@@ -0,0 +1,69 @@
...
+name        = "darwinia-ethereum-issuing"
+readme      = "README.md"
+repository  = "https://github.com/darwinia-network/darwinia-common/"
+version     = "2.1.0"
+
+[dependencies]
+# crates
+codec               = { package = "parity-scale-codec", version = "2.0.1", default-features = false, features = ["derive"] }
+ethereum-types      = { default-features = false, version = "0.11.0" }
+ethereum-primitives = { default-features = false, path = "../../../../primitives/ethereum-primitives" }
```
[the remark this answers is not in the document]
AsceticBear: Yes, we should. I will move it in the next commit.

AsceticBear (approved):

AurevoirXavier:
I'll review this later.

frame/bridge/ethereum/issuing/src/lib.rs
```diff
@@ -0,0 +1,419 @@
...
+		let account = Self::dvm_account_id();
+		let mapped_address = dvm_ethereum::Module::<T>::do_call(account, transaction)
+			.map_err(|e| -> &'static str { e.into() })?;
+		if mapped_address.len() != 32 {
+			return Err(Error::<T>::InvalidAddressLen.into());
+		}
+		Ok(H160::from_slice(&mapped_address.as_slice()[12..]))
+	}
+
+	pub fn token_registed(
```
HackFisher, line 348:
```suggestion
	pub fn token_registered(
```

frame/bridge/ethereum/issuing/src/lib.rs
```diff
@@ -0,0 +1,419 @@
...
+			let register_event = Abi::register_event();
+			let backing_event = Abi::backing_event();
+			let backing_address = EthereumBackingAddress::get();
+			let log_entry = verified_receipt
+				.logs
+				.into_iter()
+				.find(|x| {
+					x.address == backing_address &&
+						( x.topics[0] == register_event.signature()
+						  || x.topics[0] == backing_event.signature() )
```
HackFisher, line 177: Suggest to generalize this part and pass an Enum {Register, Redeem} as parameter.

More duplicate codes to an internal function.

frame/bridge/ethereum/issuing/src/lib.rs
```diff
@@ -0,0 +1,419 @@
...
+
+		fn deposit_event() = default;
+
+		fn on_initialize(_n: BlockNumber<T>) -> Weight {
+			<BurnTokenEvents<T>>::kill();
+			0
+		}
+
+		#[weight = <T as darwinia_evm::Config>::GasWeightMapping::gas_to_weight(0x100000)]
+		pub fn register_or_redeem_erc20(origin, proof: EthereumReceiptProofThing<T>) {
```
HackFisher, line 157: Split to two different dispatch call

frame/bridge/ethereum/issuing/src/lib.rs
```diff
@@ -0,0 +1,419 @@
...
+			let account = Self::dvm_account_id();
+			let basic = <T as darwinia_evm::Config>::AccountBasicMapping::account_basic(&account);
+			let substrate_account = <T as darwinia_evm::Config>::AddressMapping::into_account_id(account.clone());
+
+			<T as Config>::RingCurrency::transfer(&user, &substrate_account, T::FeeEstimate::get(), KeepAlive)?;
+			let transaction = Self::unsigned_transaction(basic.nonce, contract, input);
+			let result = dvm_ethereum::Module::<T>::do_transact(transaction, Some(account)).map_err(|e| -> &'static str {
+				log::trace!(target: "darwinia-issuing", "register_or_issuing_erc20 error {:?}", &e);
+				e.into()
+			} )?;
```
HackFisher, line 199: Review and secondary check: internal state change, question here: should increase nonce or should not?

frame/bridge/ethereum/issuing/src/lib.rs
```diff
@@ -0,0 +1,419 @@
...
+
+	/// get dvm ethereum unsigned transaction
+	pub fn unsigned_transaction(
+		nonce: U256,
+		target: H160,
+		input: Vec<u8>,
+	) -> dvm_ethereum::Transaction {
+		dvm_ethereum::Transaction {
+			nonce,
+			gas_price: T::FeeCalculator::min_gas_price(),
```
HackFisher, line 228: Could we overpass the minimum gas price check? @AsceticBear

frame/bridge/ethereum/issuing/src/lib.rs
```diff
@@ -0,0 +1,419 @@
...
+				let ethlog = Self::parse_event(backing_event, log_entry)?;
+				Self::process_token_issuing(ethlog)?
+			};
+
+			let contract = MappingFactoryAddress::get();
+			let account = Self::dvm_account_id();
+			let basic = <T as darwinia_evm::Config>::AccountBasicMapping::account_basic(&account);
+			let substrate_account = <T as darwinia_evm::Config>::AddressMapping::into_account_id(account.clone());
+
+			<T as Config>::RingCurrency::transfer(&user, &substrate_account, T::FeeEstimate::get(), KeepAlive)?;
```
HackFisher, line 194: Could we over-pass the minimum gas price check?

It is weird to transfer currency from user in this section.

frame/bridge/ethereum/issuing/src/lib.rs
```diff
@@ -0,0 +1,419 @@
...
+			let transaction = Self::unsigned_transaction(basic.nonce, contract, input);
+			let result = dvm_ethereum::Module::<T>::do_transact(transaction, Some(account)).map_err(|e| -> &'static str {
+				log::trace!(target: "darwinia-issuing", "register_or_issuing_erc20 error {:?}", &e);
+				e.into()
+			} )?;
+
+			let leaved_balance = <T as Config>::RingCurrency::free_balance(&substrate_account);
+			// we should reserve some balance to keepalive this system account
```
HackFisher, line 202: Abnormal,

Question: what will happen to substrate_account if the substrate account data is deleted? it's evm account data will always not be deleted right?

frame/bridge/ethereum/issuing/src/lib.rs
```diff
@@ -0,0 +1,419 @@
...
+			let substrate_account = <T as darwinia_evm::Config>::AddressMapping::into_account_id(account.clone());
+
+			<T as Config>::RingCurrency::transfer(&user, &substrate_account, T::FeeEstimate::get(), KeepAlive)?;
+			let transaction = Self::unsigned_transaction(basic.nonce, contract, input);
+			let result = dvm_ethereum::Module::<T>::do_transact(transaction, Some(account)).map_err(|e| -> &'static str {
+				log::trace!(target: "darwinia-issuing", "register_or_issuing_erc20 error {:?}", &e);
+				e.into()
+			} )?;
+
+			let leaved_balance = <T as Config>::RingCurrency::free_balance(&substrate_account);
```
HackFisher, line 201:
```suggestion
			let remaining_balance = <T as Config>::RingCurrency::free_balance(&substrate_account);
```

frame/bridge/ethereum/issuing/src/lib.rs
```diff
@@ -0,0 +1,419 @@
...
+			backing,
+			token_address,
+		)
+		.map_err(|_| Error::<T>::InvalidEncodeERC20)?;
+
+		log::trace!(target: "darwinia-issuing", "register fee will be delived to fee pallet {}", fee);
+		Ok(input)
+	}
+
+	fn process_token_issuing(result: EthLog) -> Result<Vec<u8>, DispatchError> {
```
HackFisher, line 296:
```suggestion
	fn abi_encode_token_redeem(result: EthLog) -> Result<Vec<u8>, DispatchError> {
```

frame/bridge/ethereum/issuing/src/lib.rs
```diff
@@ -0,0 +1,419 @@
...
+		let ethlog = Abi::parse_event(
+			log_entry.topics.into_iter().collect(),
+			log_entry.data.clone(),
+			event,
+		)
+		.map_err(|_| <Error<T>>::EthLogPF)?;
+		Ok(ethlog)
+	}
+
+	fn process_erc20_creation(
```
HackFisher, line 252:
```suggestion
	fn abi_encode_token_creation(
```

frame/bridge/ethereum/issuing/src/lib.rs
```diff
@@ -0,0 +1,419 @@
...
+				e.into()
+			} )?;
+
+			let leaved_balance = <T as Config>::RingCurrency::free_balance(&substrate_account);
+			// we should reserve some balance to keepalive this system account
+			let maxrefund = leaved_balance.saturating_sub(<T as Config>::RingCurrency::minimum_balance());
+
+			<T as Config>::RingCurrency::transfer(&substrate_account, &user, maxrefund, KeepAlive)?;
+			VerifiedIssuingProof::insert(tx_index, true);
+			Self::deposit_event(RawEvent::RegisteredOrRedeemed(user, backing_address));
```
HackFisher, line 207: Split the Events

frame/bridge/ethereum/issuing/src/lib.rs
```diff
@@ -0,0 +1,419 @@
...
+		);
+		Ok(())
+	}
+}
+
+impl<T: Config> IssuingHandler for Module<T> {
+	/// handle
+	fn handle(address: H160, caller: H160, input: &[u8]) -> DispatchResult {
+		ensure!(MappingFactoryAddress::get() == caller, <Error<T>>::AssetAR);
+		if input.len() == 3 * 32 {
```
HackFisher, line 403: Documents for these magic number

frame/dvm/src/lib.rs
```diff
@@ -351,13 +355,20 @@ impl<T: Config> Module<T> {
...
 			Error::<T>::PreLogExists,
 		);
-		let source =
-			Self::recover_signer(&transaction).ok_or_else(|| Error::<T>::InvalidSignature)?;
+
+		// For transaction from ethereum issuing pallet, the sender set None to skip the signature verification.
+		let source = sender.map_or(
+			Self::recover_signer(&transaction).ok_or_else(|| Error::<T>::InvalidSignature)?,
+			|source| source,
+		);
```
HackFisher, line 371: This implementation looks dirty.

I think we can safely remove this part of codes since these checks have already be done in `ValidateUnsigned`.

Other internal code from other pallet is relying on audit for safety.

frame/bridge/ethereum/issuing/src/lib.rs
```diff
@@ -0,0 +1,419 @@
...
+	pub fn mapped_token_address(
+		backing: EthereumAddress,
+		source: EthereumAddress,
+	) -> Result<H160, DispatchError> {
+		let factory_address = MappingFactoryAddress::get();
+		let bytes = Abi::encode_mapping_token(backing, source)
+			.map_err(|_| Error::<T>::InvalidIssuingAccount)?;
+		let transaction = Self::unsigned_transaction(U256::from(1), factory_address, bytes);
+		let account = Self::dvm_account_id();
+		let mapped_address = dvm_ethereum::Module::<T>::do_call(account, transaction)
```
HackFisher, line 340: Why we need a `do_call` here? Is it a transaction or simply a evm readonly query?

frame/bridge/ethereum/issuing/src/lib.rs
```diff
@@ -0,0 +1,419 @@
...
+}
+
+impl<T: Config> IssuingHandler for Module<T> {
+	/// handle
+	fn handle(address: H160, caller: H160, input: &[u8]) -> DispatchResult {
+		ensure!(MappingFactoryAddress::get() == caller, <Error<T>>::AssetAR);
+		if input.len() == 3 * 32 {
+			let registed_info =
+				TokenRegisterInfo::decode(input).map_err(|_| Error::<T>::InvalidInputData)?;
+			Self::token_registed(registed_info.0, registed_info.1, registed_info.2)
```
HackFisher, line 406:
```suggestion
			Self::finish_token_register(registed_info.0, registed_info.1, registed_info.2)
```

HackFisher (changes requested):

frame/bridge/ethereum/issuing/src/lib.rs
xiaoch05: here, we construct the transaction with the latest nonce each time. And the nonce is increased by `dvm` when the transaction is sent.

frame/bridge/ethereum/issuing/src/lib.rs
xiaoch05: It's not a transaction. Only a simple query method without any fees.

frame/dvm/src/lib.rs
xiaoch05: remove param sender and use address 0x0 as a runtime common caller address.

frame/bridge/ethereum/issuing/src/lib.rs
xiaoch05: use evm API directly.

frame/bridge/ethereum/issuing/src/lib.rs
xiaoch05: gas_price set to `None` to avoid gas fee problem

frame/bridge/ethereum/issuing/src/lib.rs
xiaoch05: I tried to set `nonce = None` here for this account addressed `0x0`, and there is no problem.
@AsceticBear can you recheck this case?

AsceticBear: > I tried to set nonce = None here for this account addressed 0x0, and there is no problem.

You skip the [nonce check](https://github.com/darwinia-network/darwinia-common/blob/master/frame/evm/src/runner/stack.rs#L98) in EVM runner when tx nonce is  `None`.  

The 0x0's nonce will increase by one after each `do_transact` call even if the tx nonce is None. I think we would better to use the latest account nonce for the transaction from the issuing module, just like ordinary tx.

frame/bridge/ethereum/issuing/src/lib.rs
```diff
@@ -0,0 +1,423 @@
...
+			log_entry.topics.into_iter().collect(),
+			log_entry.data.clone(),
+			log_event,
+		)
+		.map_err(|_| <Error<T>>::EthLogPF)?;
+
+		Ok((tx_index, ethlog))
+	}
+
+	pub fn call_mapping_factory(input: Vec<u8>) -> DispatchResult {
```
HackFisher, line 387:
```suggestion
	pub fn transact_mapping_factory(input: Vec<u8>) -> DispatchResult {
```

frame/bridge/ethereum/issuing/src/lib.rs
HackFisher: `T::FeeCalculator::min_gas_price()` is not None actually?  Why not directly use None here, and require document here.

frame/bridge/ethereum/issuing/src/lib.rs
```diff
@@ -0,0 +1,423 @@
...
+	pub fn unsigned_transaction(target: H160, input: Vec<u8>) -> dvm_ethereum::Transaction {
+		dvm_ethereum::Transaction {
+			nonce: U256::zero(),
+			gas_price: T::FeeCalculator::min_gas_price(),
+			gas_limit: U256::from(0x100000),
+			action: dvm_ethereum::TransactionAction::Call(target),
+			value: U256::zero(),
+			input,
+			signature: TransactionSignature::new(
+				0x78,
```
HackFisher, line 206: What's this magic number?

frame/bridge/ethereum/issuing/src/lib.rs
```diff
@@ -0,0 +1,423 @@
...
+		<BurnTokenEvents<T>>::append(system_event);
+		Self::deposit_event(raw_event);
+		T::EcdsaAuthorities::schedule_mmr_root(
+			(<frame_system::Pallet<T>>::block_number().saturated_into::<u32>() / 10 * 10 + 10)
+				.saturated_into(),
+		);
+		Ok(())
+	}
+
+	pub fn burn_token(
```
HackFisher, line 325:
```suggestion
	pub fn deposit_burn_token_event(
```

frame/dvm/src/lib.rs
```diff
@@ -351,8 +367,8 @@ impl<T: Config> Module<T> {
 			transaction.input.clone(),
 			transaction.value,
 			transaction.gas_limit,
-			Some(transaction.gas_price),
-			Some(transaction.nonce),
+			gas_price,
```
HackFisher, line 370: OK, being overwrite here.

HackFisher (changes requested):
@xiaoch05 Some more naming suggestions.

frame/dvm/src/lib.rs
```diff
@@ -420,6 +436,27 @@ impl<T: Config> Module<T> {
 		.into())
 	}
 
+	pub fn do_call(contract: H160, input: Vec<u8>) -> Result<Vec<u8>, DispatchError> {
```
AsceticBear, line 439: Why not use darwinia_evm runner call directly?

xiaoch05: almost all the contract methods the issuing pallet used comes from the `dvm pallet`.
here I want to keep the call graph simple.
but I'm not sure whether it's suitable to use darwinia_evm directly or to use dvm.
If we use evm directly here, we do not need to add `do_call` function in dvm.

AsceticBear: Please try it! No need to add this `do_all`.

```rs
pub trait dvm::Config: frame_system::Config<Hash = H256> + pallet_timestamp::Config + darwinia_evm::Config
```

AsceticBear: Sorry, I've rethought it again. Using darwinia-evm's runner call directly also requires abstraction work and is not yet as easy to understand as it is now. Let's keep the existing implementation and ignore my previous comments.

frame/bridge/ethereum/issuing/src/lib.rs
AsceticBear: The transaction defined as:

```rs
pub struct Transaction {
	pub nonce: U256,
	pub gas_price: U256,
	pub gas_limit: U256,
	pub action: TransactionAction,
	pub value: U256,
	pub input: Bytes,
	pub signature: TransactionSignature,
}
```
So, gas_price can not be None here. 

```rs
let (source, gas_price, nonce) = if need_signer {
	(
		Self::recover_signer(&transaction).ok_or_else(|| Error::<T>::InvalidSignature)?,
		Some(transaction.gas_price),
		Some(transaction.nonce),
	)
	} else {
		(H160::zero(), None, None)
        };
```
The final gas_price value will be None in `do_transact`, if I understand correctly.

xiaoch05: so, it's better to set gas_price as zero here.

frame/bridge/ethereum/issuing/src/lib.rs
```diff
@@ -0,0 +1,424 @@
...
+		pub BurnTokenEvents get(fn burn_token_events): Vec<<T as frame_system::Config>::Event>;
+	}
+}
+
+decl_module! {
+	pub struct Module<T: Config> for enum Call
+	where
+		origin: T::Origin
+	{
+		const FeeEstimate: RingBalance<T> = T::FeeEstimate::get();
```
AurevoirXavier, line 149: Any usage of `FeeEstimate`?

xiaoch05: this should be deleted.

AurevoirXavier: Okay, I'll delete this directly.

frame/bridge/ethereum/issuing/src/lib.rs
```diff
@@ -0,0 +1,423 @@
...
+		fn deposit_event() = default;
+
+		fn on_initialize(_n: BlockNumber<T>) -> Weight {
+			<BurnTokenEvents<T>>::kill();
+			0
+		}
+
+		#[weight = <T as darwinia_evm::Config>::GasWeightMapping::gas_to_weight(0x100000)]
+		pub fn register_erc20(origin, proof: EthereumReceiptProofThing<T>) {
+			log::info!(target: "darwinia-issuing", "start to register erc20 token");
```
AurevoirXavier, line 160: log will target the crate's name automactically

AurevoirXavier: Also you use info here. These log will always be printed.

xiaoch05: forgot to remove all these logs used to debug.

frame/bridge/ethereum/issuing/src/lib.rs
```diff
@@ -0,0 +1,423 @@
...
+		)
+		.map_err(|_| <Error<T>>::EthLogPF)?;
+
+		Ok((tx_index, ethlog))
+	}
+
+	pub fn transact_mapping_factory(input: Vec<u8>) -> DispatchResult {
+		let contract = MappingFactoryAddress::get();
+		let transaction = Self::unsigned_transaction(contract, input);
+		let result = dvm_ethereum::Module::<T>::do_transact(transaction, false).map_err(
```
HackFisher, line 385: Suggest to remove the unsigned_transaction here, and using a new internal call `mock_transact(input: Vec<u8>)`

frame/bridge/ethereum/issuing/src/lib.rs
```diff
@@ -0,0 +1,423 @@
...
+impl<T: Config> Module<T> {
+	/// The account ID of the issuing pot.
+	pub fn dvm_account_id() -> H160 {
+		let account32: AccountId32 = T::ModuleId::get().into_account();
+		let account20: &[u8] = &account32.as_ref();
+		H160::from_slice(&account20[..20])
+	}
+
+	/// get dvm ethereum unsigned transaction
+	pub fn unsigned_transaction(target: H160, input: Vec<u8>) -> dvm_ethereum::Transaction {
```
HackFisher, line 195: Remove this. And move to `mock_transact(input: Vec<u8>)`

frame/dvm/src/lib.rs
```diff
@@ -334,13 +338,25 @@ impl<T: Config> Module<T> {
 		}
 	}
 
-	fn do_transact(transaction: ethereum::Transaction) -> DispatchResultWithPostInfo {
+	pub fn do_transact(
+		transaction: ethereum::Transaction,
+		need_signer: bool,
+	) -> DispatchResultWithPostInfo {
```
HackFisher, line 344: Do not change this and use a new `internal_transact` instead.

HackFisher (changes requested):

frame/bridge/ethereum/issuing/src/lib.rs
```diff
@@ -0,0 +1,417 @@
...
```
HackFisher, line 195:
```suggestion
	pub fn internal_transaction(target: H160, input: Vec<u8>) -> dvm_ethereum::Transaction {
```

HackFisher: Even though this is an internal transaction, it will still be stored in ethereum block.

So we need to mock a transaction anyway.

AurevoirXavier (commented):
LGTM.

frame/dvm/src/lib.rs
```diff
@@ -81,6 +81,52 @@ impl Default for EthereumStorageSchema {
 	}
 }
 
+/// The ethereum transaction include recovered source account
+pub struct EthereumTransaction {
```
AsceticBear, line 85: Two suggestions here:
1. We can move this struct at the end of  https://github.com/darwinia-network/darwinia-common/blob/master/primitives/rpc/src/transaction.rs.
2. `EthereumTransaction` rename to `InternalTransaction` would be better.

HackFisher: I found that `EthereumTransaction` is used temporally in memory in this file, and also used by normal transaction.

I'm changing it to DVMTransaction.

xiaoch05: Yes, it's used to convert the normal transaction to the format `execute` fn used.

HackFisher (approved):

AurevoirXavier (approved):

PR #549, opened 2021-03-18
Fix transfer gas price

AsceticBear:
image

### DIFF
--- a/bin/node/runtime/pangolin/src/evm.rs
+++ b/bin/node/runtime/pangolin/src/evm.rs
@@ -8,12 +8,12 @@ use darwinia_evm::{
 use darwinia_evm_precompile::DarwiniaPrecompiles;
 use dvm_ethereum::account_basic::DVMAccountBasicMapping;
 
-/// Fixed gas price of `1`.
+/// Fixed gas price.
 pub struct FixedGasPrice;
 impl FeeCalculator for FixedGasPrice {
 	fn min_gas_price() -> U256 {
 		// Gas price is always one token per gas.
-		1.into()
+		1_000_000_000.into()
 	}
 }
 frame_support::parameter_types! {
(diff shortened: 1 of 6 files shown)


bin/node/runtime/pangolin/src/evm.rs
```diff
@@ -8,12 +8,12 @@ use darwinia_evm::{
...
 use dvm_ethereum::account_basic::DVMAccountBasicMapping;
 
-/// Fixed gas price of `1`.
+/// Fixed gas price.
 pub struct FixedGasPrice;
 impl FeeCalculator for FixedGasPrice {
 	fn min_gas_price() -> U256 {
 		// Gas price is always one token per gas.
-		1.into()
+		100_000_000_000u128.into()
```
AsceticBear, line 16: About this value, I would like to hear your advice. @HackFisher

HackFisher: @AsceticBear What's the meaning of minimal gas price, is this some fixed gas price which can not be changed by user?

What will happen if user manually change it to lower gas price, e.g. 1 Gwei, in Metamask?

HackFisher (changes requested):

AsceticBear:
> What's the meaning of minimal gas price, is this some fixed gas price which can not be changed by user?

Yes, min_gas_price represents the minimum value set for the transaction gas_price. The user can set a gas_price greater than this value. However, if it is less than this value, the transaction will be rejected by the transaction pool. More info https://github.com/darwinia-network/darwinia-common/blob/master/frame/dvm/src/lib.rs#L305

HackFisher:
Currently,

Frontier and DVM are using `FeeCalculator::min_gas_price()` to response `gas_price` rpc call which is a arbitrary way, this might due the incomplete dynamic-fee part in ethereum mempool.

```
fn gas_price() -> U256 {
	<Runtime as pallet_evm::Config>::FeeCalculator::min_gas_price()
}
```

Since there are enough block gas limit(? to check), we can use this `min_gas_price` for now, it should be enough for inclusion.

To make Metamask show non-zero gas price, I would suggest to use 1 Gwei. `1_000_000_000u128.into()`

HackFisher (approved):

PR #554, opened 2021-03-19
More tests about account basic mapping

AsceticBear:


frame/dvm/src/account_basic.rs
```diff
@@ -6,9 +6,11 @@ use sp_runtime::{
 	SaturatedConversion,
 };
 
+type RingInstance = darwinia_balances::Instance0;
+
 pub struct DVMAccountBasicMapping<T>(sp_std::marker::PhantomData<T>);
 
-impl<T: crate::Config + darwinia_balances::Config<darwinia_balances::Instance0>> AccountBasicMapping
+impl<T: crate::Config + darwinia_balances::Config<RingInstance>> AccountBasicMapping
```
AurevoirXavier, line 13: How about

```
pub struct DVMAccountBasicMapping<T, I>(sp_std::marker::PhantomData<T, I>);

impl<T: crate::Config + darwinia_balances::Config<I>, I> AccountBasicMapping for DVMAccountBasicMapping<T, I> {
    // ...
}
```

HackFisher:
@AsceticBear File conflicts need to be resolved before continue reviewing and merge.

frame/dvm/src/account_basic.rs
AsceticBear: I tried this approach yesterday, see https://github.com/AsceticBear/darwinia-common/commit/babe72e9f3bdac298256e8b5349fc132cb5b24b8

However, I decided to abandon this approach for the following reasons.
1. this implementation will bring data compatibility problems, now the DVM API query is based on storage override, after adding Instance, the query will fail with the old override.
2. there is no strong demand for DVM to implement multiple instances now.

HackFisher: > I tried this approach yesterday, see [AsceticBear@babe72e](https://github.com/AsceticBear/darwinia-common/commit/babe72e9f3bdac298256e8b5349fc132cb5b24b8)
> 
> However, I decided to abandon this approach for the following reasons.
> 
> 1. this implementation will bring data compatibility problems, now the DVM API query is based on storage override, after adding Instance, the query will fail with the old override.
> 2. there is no strong demand for DVM to implement multiple instances now.

Yes, it makes sense, especially  we have no demand for multiple instances and simple and moderate abstraction is good.

AurevoirXavier: Actually, I've a question.

Why we need this bound? `trait Config: darwinia_balance<RingInstance>`

Can't find any storage usage here. Only the type. `T::Balance`

But the best practice for define this type is `Config::RingCurrency::Balance`

AurevoirXavier:
Take a look at the latest commit.

If not good just revert it.

AsceticBear:
> Take a look at the latest commit.
> 
> If not good just revert it. ; )

I like these commit changes. LGTM

AurevoirXavier (approved):

HackFisher (approved):

PR #558, opened 2021-03-22
More Genesis Validators

AurevoirXavier:
Update version to 2.1.0, runtime 210

Increase genesis validator from  2 to 6

### DIFF
--- a/.maintain/insert-session-keys.sh
+++ b/.maintain/insert-session-keys.sh
@@ -1,5 +1,9 @@
 #!/bin/sh
 
+# $1 is uri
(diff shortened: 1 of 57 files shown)


.maintain/insert-session-keys.sh
```diff
@@ -0,0 +1,56 @@
+#!/bin/sh
+
```
AsceticBear, line 2: Maybe we should add some documentation about what `$2, $3` stands for.  It would be nice to have an example here.

HackFisher (approved):

PR #566, opened 2021-03-25
Substrate Updates

AurevoirXavier:
https://github.com/darwinia-network/substrate-update-tracking/issues/7


bin/node/runtime/pangolin/src/lib.rs
```diff
@@ -421,78 +428,78 @@ frame_support::construct_runtime! {
...
 	{
 		// Basic stuff; balances is uncallable initially.
-		System: frame_system::{Module, Call, Storage, Config, Event<T>} = 0,
-		RandomnessCollectiveFlip: pallet_randomness_collective_flip::{Module, Call, Storage} = 1,
+		System: frame_system::{Pallet, Call, Storage, Config, Event<T>} = 0,
+		RandomnessCollectiveFlip: pallet_randomness_collective_flip::{Pallet, Call, Storage} = 1,
 
 		// Must be before session.
-		Babe: pallet_babe::{Module, Call, Storage, Config, ValidateUnsigned} = 2,
+		Babe: pallet_babe::{Pallet, Call, Storage, Config, ValidateUnsigned} = 2,
```
HackFisher, line 435: We need to double check `Babe` storage prefix if it does pallet macro syntax migration

frame/staking/src/lib.rs
```diff
@@ -286,9 +283,83 @@ pub mod weights;
...
+			log!(info, "QueuedScore.exits()? {:?}", QueuedScore::exists());
+			// these must exist.
+			assert!(
+				IsCurrentSessionFinal::exists(),
+				"IsCurrentSessionFinal storage item not found!"
+			);
+			assert!(
+				EraElectionStatus::exists(),
+				"EraElectionStatus storage item not found!"
+			);
```
HackFisher, line 341: What will happen if these `assert!` fail.

frame/staking/src/lib.rs
```diff
@@ -286,9 +283,83 @@ pub mod weights;
...
+		use super::*;
+		use frame_support::{pallet_prelude::*, traits::Get, weights::Weight};
+
+		macro_rules! generate_storage_types {
+			($name:ident => Value<$value:ty>) => {
+				paste::paste! {
+					struct [<$name Instance>];
+					impl frame_support::traits::StorageInstance for [<$name Instance>] {
+						fn pallet_prefix() -> &'static str {
+							"DarwiniaStaking"
```
HackFisher, line 301: Double check

frame/staking/src/lib.rs
```diff
@@ -892,44 +884,8 @@ decl_storage! {
...
-		/// TWO_PHASE_NOTE: should be removed once we switch to multi-phase.
-		pub EraElectionStatus get(fn era_election_status): ElectionStatus<T::BlockNumber>;
-
-		/// True if the current **planned** session is final. Note that this does not take era
-		/// forcing into account.
-		///
-		/// TWO_PHASE_NOTE: should be removed once we switch to multi-phase.
-		pub IsCurrentSessionFinal get(fn is_current_session_final): bool = false;
-
 		/// This is set to v5.0.0 for new networks.
```
HackFisher, line 887: Update comment

frame/staking/src/lib.rs
```diff
@@ -3341,14 +2682,15 @@ impl<T: Config> Module<T> {
 				let mut total_power = 0;
 				let mut others = Vec::with_capacity(support.voters.len());
 				support.voters.into_iter().for_each(|(nominator, weight)| {
-					let origin_weight = Self::power_of(&nominator) as ExtendedBalance;
+					let origin_weight = Self::power_of(&nominator) as Balance;
```
HackFisher, line 2685: What's the difference between `ExtendedBalance` and `Balance`, just renaming.

`Balance` looks like being the type of Currency.

HackFisher (changes requested):

frame/staking/src/lib.rs
AurevoirXavier: Both of these 2 type are `u128`

We use it for calculate our vote weight.

frame/staking/src/lib.rs
AurevoirXavier: Will be test on Pangolin first

bin/node/runtime/pangolin/src/lib.rs
AurevoirXavier: Just a code level rename.

https://github.com/paritytech/substrate/pull/8372

frame/staking/src/lib.rs
HackFisher: I suggest to continue use ExtendedBalance.

Even though they are both u128, but for the weight in `support.voters` it is using `npos_election::ExtendedBalance`, we should use the same from the point of code readability and type system.

AurevoirXavier: Actually they use `Balance` as `Vote` now.

https://sourcegraph.com/github.com/paritytech/substrate/-/blob/frame/staking/src/lib.rs#L2366

`to_currency` retuan a `Balance` type.

I'd like to invoke a new type `Vote = u128` here, to clarify the usage.

HackFisher: Substrate use `VoteWeight`(u64) and `ExtendedBalance`(u128) to represent vote in the election.

Here is an convert from `VoteWeight` to `ExtendedBalance`

https://github.com/paritytech/substrate/blob/2e34d473e6ba81b454eb039544f2162975c0b30b/primitives/npos-elections/src/helpers.rs#L38

Substrate's staking is more simple because they can directly use Currency as Vote, but here in our case, we should use `Power` (combination of ratio power from ring and kton).

When converting `ExtendedBalance` back to `Power` (from u128 to u32), we should have a `to_power` just like substrate's `to_currency` to strictly avoid overflow.

Anyway, I don't think we should use `Balance` here,  as we are using different thing as vote compared to substrate staking.

AurevoirXavier: We don't need to convert vote back to currency.

We record all balance value in the exposure which use for slashing.

Actually our power is the vote here.

I think we just need to alias `power` as `vote` (a `u128` type power).

HackFisher: > We don't need to convert vote back to currency.
> 
> We record all balance value in the exposure which use for slashing.
> 
> Actually our power is the vote here.
> 
> I think we just need to alias `power` as `vote` (a `u128` type `power`).
> 
> ```rust
> // a `u128` type `Power`
> type Vote = u128;
> ```

Not convert vote back to currency, need convert vote back to power.

The purpose of this type converter is pure for strict overflow check.

AurevoirXavier: Could you point out which line we convert the `u128` back to `u32`?

Let me check.

HackFisher: https://github.com/darwinia-network/darwinia-common/blob/a00d1031acb941161d79d9d06611a3b27539b60f/frame/staking/src/lib.rs#L3379

frame/staking/src/lib.rs
AurevoirXavier: This is a `fn pre_migrate`

And this only work when you use the `try-runtime` subcommand

HackFisher (approved):

Issue #561, opened 2021-03-23
Migrate All Pallets to Attribute Macro


songtianyi:
Here is the list of pallet and their status, if you want to migrate one you can put your name after it to avoid duplicated work.
some helpful guidelines to do the migration: https://crates.parity.io/frame_support/attr.pallet.html#upgrade-guidelines
some helpful guidelines to review a migration: https://crates.parity.io/frame_support/attr.pallet.html#checking-upgrade-guidelines

- [x] darwinia-crab-backing @songtianyi #560 
- [x] darwinia-crab-issuing @songtianyi #568 
- [x] darwinia-balances @AurevoirXavier #593 
- [x] darwinia-tron-backing @AurevoirXavier #610
- [x] darwinia-vesting @AurevoirXavier #586
- [x] darwinia-ethereum-backing @AurevoirXavier #611
- [ ] darwinia-ethereum-issuing @AurevoirXavier 
- [ ] darwinia-relayer-game
- [ ] darwinia-relay-authorities
- [x] darwinia-staking @AurevoirXavier #924
- [x] darwinia-header-mmr @AurevoirXavier #613
- [x] darwinia-evm @AsceticBear  #648 
- [x] dvm-dynamic-fee @AurevoirXavier @AsceticBear #644
- [x] dvm-ethereum @AsceticBear #649

AurevoirXavier:
And I propose this style:

```rust
// LICENSE

//! # Module Introduction

//(ignore this in real code) pub import don't need comment
pub use pallet::*;
pub use weights::WeightInfo;

//(ignore this in real code) test module above others
#[cfg(test)]
mod mock;
#[cfg(test)]
mod tests;

#[frame_support::pallet]
pub mod pallet {
	//(ignore this in real code) pub mod types go first
	pub mod types {
		//(ignore this in real code) all import need a comment, easy for code review
		// --- darwinia ---
		use super::*;

		// Generic type
		pub type AccountId<T> = <T as frame_system::Config>::AccountId;
		// Simple type
		pub type SimpleNum = u128;
	}
	pub use types::*;

	//(ignore this in real code) imports go second
	//(ignore this in real code) all import need a comment, easy for code review
	// --- substrate ---
	use frame_support::pallet_prelude::*;
	use frame_system::pallet_prelude::*;
	//(ignore this in real code) all import need a comment, easy for code review
	// --- darwinia ---
	use crate::weights::WeightInfo;

	//(ignore this in real code) trait goes third
[...]

PR #568, opened 2021-03-26
feat: migrate crab issuing to use pallet!

songtianyi:
Part of #561.

Converts the crab backing pallet to the new pallet attribute macro introduced in [#6877](https://github.com/paritytech/substrate/pull/6877).

[Upgrade guidelines used](https://substrate.dev/rustdocs/v3.0.0/frame_support/attr.pallet.html#upgrade-guidelines).

⚠️ Breaking Change ⚠️
From [checking upgrade guidelines](https://crates.parity.io/frame_support/attr.pallet.html#checking-upgrade-guidelines)

> storages now use PalletInfo for module_prefix instead of the one given to decl_storage: use of this pallet in construct_runtime! needs careful updating of the name in order to not break storage or to upgrade storage (moreover for instantiable pallet). If pallet is published, make sure to warn about this breaking change.

Common notes
- their Trait is renamed Config
- all storages are now generic over T
- module prefix used by storages is now the PalletInfo::name given by construct_runtime.
- no longer RawEvent.

The metadata for CrabIssuing before migration
```json
  {
          name: CrabIssuing,
          storage: {
            prefix: DarwiniaCrabIssuing,
            items: [
              {
                name: TotalMappedRing,
                modifier: Default,
                type: {
                  Plain: MappedRing
                },
                fallback: 0x00000000000000000000000000000000,
                documentation: []
              }
            ]
          },
          calls: [],
          events: [
            {
              name: DummyEvent,
              args: [
                AccountId,
                RingBalance,
                MappedRing
              ],
              documentation: [
                 Dummy Event. [who, swapped *CRING*, burned Mapped *RING*]
              ]
            }
          ],
          constants: [
            {
              name: ModuleId,
              type: ModuleId,
              value: da/crais,
              documentation: []
            }
          ],
          errors: [],
          index: 33
        },
```

The metadata for CrabIssuing after migration
```json
 {
          name: DarwiniaCrabIssuing,
          storage: {
            prefix: DarwiniaCrabIssuing,
            items: [
              {
                name: TotalMappedRing,
                modifier: Optional,
                type: {
                  Plain: MappedRing
                },
                fallback: 0x00,
                documentation: []
              }
            ]
          },
          calls: [],
          events: [
            {
              name: DummyEvent,
              args: [
                AccountId,
                RingBalance,
                MappedRing
              ],
              documentation: [
                 Dummy Event. [who, swapped *CRING*, burned Mapped *RING*]
              ]
            }
          ],
          constants: [
            {
              name: ModuleId,
              type: ModuleId,
              value: da/crais,
              documentation: []
            }
          ],
          errors: [],
          index: 33
        },
```


frame/bridge/crab/issuing/src/lib.rs
```diff
@@ -42,73 +46,76 @@ mod types {
...
+	#[pallet::hooks]
+	impl<T: Config> Hooks<BlockNumberFor<T>> for Pallet<T> {}
+
+	#[pallet::call]
+	impl<T: Config> Pallet<T> {}
+
+	#[pallet::genesis_config]
+	pub struct GenesisConfig<T> {
+		pub total_mapped_ring: MappedRing,
+		pub phantom: std::marker::PhantomData<T>,
```
songtianyi, line 97: To be compatible with mock, added a phantom here, not sure it's a good option, any suggestions? @AurevoirXavier 
Because the following code builds storage with a type Test, so the GenesisConfig needs to be defined with a T which not used actually. To avoid a build error `unused generic T xxx`, the PhantomData was added.

```rust
        let mut t = frame_system::GenesisConfig::default()
                .build_storage::<Test>()
                .unwrap();
```

AurevoirXavier: We don't need the generic config.

Just keep it simple/

AurevoirXavier: I've fixed that.

HackFisher (approved):

AurevoirXavier (approved):

Issue #575, opened 2021-04-02
KTON_PROCOMPILE issues


hujw77:
Function `transfer_and_call(address wkton_addr_,uint256 value_)`

### Function selector params issue.
`transfer_and_call(address,address,uint256)`  ==> `transfer_and_call(address,uint256)`

The function only use two paramters, but the sig is three.

### Transfer decimal issue.
`transfer_and_call(address,uint256)`
The second parameter `value_` decimal should be the same as `PKTON`, maybe `9`

When I transfer 1 `PKTON`, the `value_` should be `1 * 1e9`, but now is `1`.

### Transfer succeed when `wkton_addr_` is KTON_PROCOMPILE itself or any other address.
When do `wkton_addr_.deposit()`. If the call is failed, the transaction should revert. 

#### Script
```shell
curl -fsS https://pangolin-rpc.darwinia.network -d '{"id":1,"jsonrpc":"2.0","method":"eth_estimateGas","params":[{"data":"0x784deab50000000000000000000000000000000000000000000000000000000000000016000000000000000000000000000000000000000000000000000000000000000a","from":"0x0f14341A7f464320319025540E8Fe48Ad0fe5aec","gas":"0x5b8d80","to":"0x0000000000000000000000000000000000000016"}]}' -H 'Content-Type: application/json'
```
#### Node info
```shell
2021-04-02 15:12:50.486 DEBUG        http.worker00 evm: Execution Error(Other("Invalid Action！")) [source: 0x0000000000000000000000000000000000000016, value: 0, gas_limit: 999978656, actual_fee: 0]
2021-04-02 15:12:50.487 DEBUG        http.worker00 evm: Execution Succeed(Returned) [source: 0x0f14341a7f464320319025540e8fe48ad0fe5aec, value: 0, gas_limit: 1000000000, actual_fee: 0]
2021-04-02 15:12:50.487 DEBUG        http.worker00 evm: Execution Error(Other("Invalid Action！")) [source: 0x0000000000000000000000000000000000000016, value: 0, gas_limit: 499989156, actual_fee: 0]
2021-04-02 15:12:50.487 DEBUG        http.worker00 evm: Execution Succeed(Returned) [source: 0x0f14341a7f464320319025540e8fe48ad0fe5aec, value: 0, gas_limit: 500010500, actual_fee: 0]
```

HackFisher:
Need more description about what the issues is and what is the expected behavior.

HackFisher:
[...]

PR #577, opened 2021-04-06
Generic account basic

AsceticBear:
This PR is the first step for #575. It would be much clear to hack precompile after this generic refactor.


HackFisher (approved):

AsceticBear:
@AurevoirXavier  Need your review also.

AurevoirXavier:
> @AurevoirXavier Need your review also.

Ing

AurevoirXavier (changes requested):
The implementation looks not good. Too many dup def.

I'm going to refactor this.

E.g

```
type RingCurrency = Ring;
type RingAccountBasic = DvmAccountBasic<Self, Ring, RingRemainBalance>;
                                              ^^^^ just Self::RingCurrency                                    
```

AsceticBear:
> I'm going to refactor this.

Go and make it better.

AurevoirXavier:
After I reading evm codes. I found it's not easy to apply the refactor. And don't want to make too much breaking changes.

Do you think is good to use a custom `Basic` instead evm's?

```rust
Basic {
	balance,
	nonce,
}

// ---

Basic {
	ring,
	kton,
	nonce,
}
```

AurevoirXavier (approved):

AsceticBear:
> After I reading the evm codes. I found it's not easy to apply the refactoring. And I don't want to introduce too much breaking changes.
> 
> BTW, do you think is good to use a custom `Basic` instead evm's?
> 
> ```rust
> Basic {
> 	balance,
> 	nonce,
> }
> 
> // ---
> 
> Basic {
> 	ring,
> 	kton,
> 	nonce,
> }
> ```

I do not think this is a good idea. Because only `RING` will be consumed during evm execution.

The `KtonAccountBasic` is designed to purely handle issues about Kton precompile.

PR #579, opened 2021-04-07
Common node without DVM [DO NOT MERGE]

AsceticBear:
This is a branch without DVM for testing integrate DVM to the Crab network.

This draft pull request is intended to give a clearer picture of what the DVM consists of. So, no need to merge.


bin/node/cli/src/chain_spec.rs
```diff
@@ -252,6 +252,7 @@ fn pangolin_build_spec_genesis() -> pangolin_runtime::GenesisConfig {
 			balances: vec![
 				(root.clone(), BUNCH_OF_COINS),
 				(get_account_id_from_seed::<sr25519::Public>("Alice"), A_FEW_COINS),
+				(get_account_id_from_seed::<sr25519::Public>("Bob"), A_FEW_COINS),
```
AsceticBear, line 255: Change genesis config to avoid sync from pangolin node in my local test env. Just ignore it.

PR #580, opened 2021-04-07
Common node with DVM [DO NOT MERGE]

AsceticBear:
The intention of this pr is to give an idea of the extra work we need to do based on the darwinia-common master in order to integrate DVM into crab.

The commit-msg is a litter mess now. I will make it clear later.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5383,6 +5383,7 @@ dependencies = [
  "darwinia-treasury",
  "darwinia-tron-backing",
  "darwinia-vesting",
+ "dp-storage",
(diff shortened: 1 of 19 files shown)


HackFisher:
I think we are good to go in this direction by removing FrontierBlockImport.

AsceticBear:
> I think we are good to go in this direction by removing FrontierBlockImport.

I have tests the case of using `FrontierBlockImport` and no `FrontierBlockImport`. That does not affect node upgrade. I prefer use original grandpa import.

Issue #578, opened 2021-04-07
KTON precompile bug


HackFisher:
Related #575 

https://github.com/darwinia-network/darwinia-common/blob/66ec3fb06b9ff3850fa3c909cc0150c3514808bc/frame/evm/precompile/contracts/kton/src/lib.rs#L79

There are bugs related to calculation for `transfer_value `.

AsceticBear:
This will be fixed in the KTON precompile refactor pr later which i am working on.

PR #581, opened 2021-04-08
Refactor kton precompile

### LINKED ISSUE #575: KTON_PROCOMPILE issues

AsceticBear:
This pr fix #575 #578 and implemented based on #577. Although this pr must wait for #577 merged. It is ready for reviewing now.

This precompile is **critical** to the whole system, so please make sure you spend some time reviewing it.

@HackFisher @hujw77 @AurevoirXavier 

---

With regard to the design of the KTON precompile, I think it necessary to write the following material to help those who are not familiar with this precompile contract to understand it.

Precompile contracts are special contracts that have a fixed address and the user can send transactions to call methods in the contract. The KTON contract we are discussing is a pre-compiled contract with a fixed address of `0000000000000000000000000000000000000016`.

The purpose of the KTON precompile is to support the transfer of the native token KTON across the Darwinia network and the DVM warped contract like [WCKTON Contract](https://github.com/evolutionlandorg/evo-deploy/blob/pgl/src/wrapper/WCKTON.sol).

Details implementation;

There are two actions designed in KTON precompile:

1. Transfer and call (Darwinia native kton -> wkton contract token)

When using this action, we call the precompile directly. The precompile input includes two-part:

```rs
pub struct CallData {
	wkton_address: H160,
	value: U256,
}
```

Once we receive call from outside and parse the `call_data`, we transfer kton from caller to wkton contract firstly, then call ` deposit(address from, uint256 value)` in WCKTON Contract which deployed in advance.

2. Withdraw (wkton contract token -> Darwinia native kton)

We directly call WCKTON Contract `withdraw(bytes32 to, uint wad)`, this method will make a call inside to the KTON precompile withdraw action.

The input includes two-parts also:

```rs
#[derive(Debug, PartialEq, Eq)]
pub struct WithdrawData<T: frame_system::Config> {
	pub to_account_id: AccountId<T>,
	pub kton_value: U256,
}
```
Once we receive this call and parse the withdrawal data. we transfer kton from wkton contract to the `to_account_id`.

### DIFF
--- a/frame/evm/precompile/contracts/kton/src/lib.rs
+++ b/frame/evm/precompile/contracts/kton/src/lib.rs
@@ -27,12 +27,8 @@ use evm::{Context, ExitError, ExitSucceed};
 use frame_support::{ensure, traits::Currency};
 use sha3::Digest;
 use sp_core::{H160, U256};
-use sp_runtime::traits::UniqueSaturatedInto;
-use sp_runtime::SaturatedConversion;
-use sp_std::borrow::ToOwned;
-use sp_std::marker::PhantomData;
-use sp_std::prelude::*;
-use sp_std::vec::Vec;
+use sp_runtime::{traits::UniqueSaturatedInto, SaturatedConversion};
+use sp_std::{borrow::ToOwned, marker::PhantomData, prelude::*, vec::Vec};
 
 use darwinia_evm::{Account, AccountBasic, Config, Module, Runner};
 use dp_evm::Precompile;
--- a/frame/evm/precompile/contracts/kton/src/lib.rs
+++ b/frame/evm/precompile/contracts/kton/src/lib.rs
@@ -24,20 +24,19 @@ use codec::Decode;
 use core::str::FromStr;
 use ethabi::{Function, Param, ParamType, Token};
 use evm::{Context, ExitError, ExitSucceed};
-use frame_support::{
-	ensure,
-	traits::{Currency, ExistenceRequirement},
-};
+use frame_support::{ensure, traits::Currency};
 use sha3::Digest;
 use sp_core::{H160, U256};
 use sp_runtime::traits::UniqueSaturatedInto;
+use sp_runtime::SaturatedConversion;
 use sp_std::borrow::ToOwned;
 use sp_std::marker::PhantomData;
 use sp_std::prelude::*;
 use sp_std::vec::Vec;
 
-use darwinia_evm::{AddressMapping, Config, Runner};
+use darwinia_evm::{Account, AccountBasic, Config, Module, Runner};
 use dp_evm::Precompile;
+use dvm_ethereum::{account_basic::RemainBalanceOp, KtonBalance, KtonRemainBalance};
 
 type AccountId<T> = <T as frame_system::Config>::AccountId;
 
@@ -51,7 +50,7 @@ pub struct Kton<T: Config> {
 	_maker: PhantomData<T>,
 }
 
-impl<T: Config> Precompile for Kton<T> {
+impl<T: Config + dvm_ethereum::Config> Precompile for Kton<T> {
 	/// There are two actions, one is `transfer_and_call` and the other is `withdraw`
 	/// 1. Transfer_and_call Action, triggered by the user sending a transaction to the kton precompile
 	/// 	   special evm address, eg(0000000000000000000000000000000000000016). and transfer the sender's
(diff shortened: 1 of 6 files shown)


frame/evm/precompile/contracts/kton/src/lib.rs
```diff
@@ -71,31 +66,38 @@ impl<T: Config> Precompile for Kton<T> {
 			.checked_pow(U256::from(9))
 			.unwrap_or(U256::MAX);
 		let action = which_action::<T>(&input)?;
-		let con_caller = T::AddressMapping::into_account_id(context.caller);
+		// let con_caller = T::AddressMapping::into_account_id(context.caller);
```
songtianyi, line 69: commented code

frame/evm/precompile/contracts/kton/src/lib.rs
```diff
@@ -108,14 +110,39 @@ impl<T: Config> Precompile for Kton<T> {
...
-					&con_caller,
+				// Ensure wkton is a contract
+				ensure!(
+					!crate::Module::<T>::is_contract_code_empty(&context.caller),
+					ExitError::Other("The caller must be wkton contract!".into())
+				);
+				// Ensure context's apparent_value is zero
+				ensure!(
+					context.apparent_value == U256::zero(),
+					ExitError::Other("The value in tx must be zero!".into())
```
songtianyi, line 121: The apparent value in tx

frame/evm/precompile/contracts/kton/src/lib.rs
```diff
@@ -108,14 +110,39 @@ impl<T: Config> Precompile for Kton<T> {
...
+				);
+				// Ensure caller's balance is enough
+				let caller_kton = T::KtonAccountBasic::account_basic(&context.caller);
+				ensure!(caller_kton.balance >= wd.kton_value, ExitError::OutOfFund);
```
songtianyi, line 125: `OutOfFund` - we should use the same error name for the insufficient balance situation. I didn't find any other name like this by searching this repo.  
BTW we should group errors in one file/crate i think

frame/evm/src/lib.rs
```diff
@@ -563,6 +563,11 @@ impl<T: Config> Module<T> {
 		account.nonce == U256::zero() && account.balance == U256::zero() && code_len == 0
 	}
 
+	pub fn is_contract_code_empty(address: &H160) -> bool {
```
songtianyi, line 566: is_contract_code_empty(address: &H160) -> 

empty(address: &H160) or empty_address(address: &H160)

It seems this function is validating address not the contract

frame/evm/precompile/contracts/kton/src/lib.rs
songtianyi: I don't know what apparent actually means here, probably i'm wrong

AsceticBear: The `apparent` refers to tx value. here refers to context inner field from EVM executor.

frame/evm/precompile/contracts/kton/src/lib.rs
AsceticBear: The return error and `execute` function signature defined in the EVM crate. Thus you can not find it in this repo.

AurevoirXavier: > `OutOfFund` - we should use the same error name for the insufficient balance situation. I didn't find any other name like this by searching this repo.
> BTW we should group errors in one file/crate i think

No you can't.

Each error belong to its `mod pallets;`

frame/evm/src/lib.rs
AsceticBear: In the Ethereum system, an address can represent both a user account and a contract account. The difference between them is whether they have contract code or not. Here is represent contract account.

frame/evm/precompile/contracts/kton/src/lib.rs
songtianyi: Yes, but we should have a name convention for error, event names, etc

frame/evm/src/lib.rs
```diff
@@ -340,6 +336,10 @@ decl_storage! {
 					balance: account.balance,
 					nonce: account.nonce,
 				});
+				T::KtonAccountBasic::mutate_account_basic(&address, Account {
+					balance: account.balance,
+					nonce: account.nonce,
+				});
```
AsceticBear, line 342: Give some kton to init evm account for js test.

HackFisher (approved):

PR #593, opened 2021-04-19
Migrate `balances` to Attribute Macro

AurevoirXavier:


AurevoirXavier:
```
2021-04-20 01:53:16 `pallet_timestamp::UnixTime::now` is called at genesis, invalid value returned: 0    
2021-04-20 01:53:16 [#0] 🗳  Finalized election round with compute ElectionCompute::OnChain.    
2021-04-20 01:53:16 [0] 💸 new validator set of size 1 has been processed for era 1    
2021-04-20 01:53:16 [darwinia-claims] Genesis Claims List is Set to EMPTY    
2021-04-20 01:53:16 initializing remote client to "http://pangolin-rpc.darwinia.network"    
2021-04-20 01:53:16 scraping keypairs from remote node http://pangolin-rpc.darwinia.network @ 0x04f3476f6cc491529ac61e8e482187b4d231ef8237b92ecff410599037800e5a    
2021-04-20 01:53:16 downloading data for all modules.    
2021-04-20 01:53:19 extending externalities with 1 manually injected keys    
2021-04-20 01:53:19 injecting a total of 7285 keys    
2021-04-20 01:53:20 Migrating `Instance0Darwinia` to `Ring`...    
2021-04-20 01:53:20 Migrating `Instance1Darwinia` to `Kton`...    
2021-04-20 01:53:20 try-runtime executed without errors. Consumed weight = 1107200000000, total weight = 512000000000000 (0.0021625) 
```

AurevoirXavier:
s2s branch require this update.

frame/balances/src/lib.rs
```diff
@@ -492,1058 +628,998 @@ decl_module! {
...
-		if let Some(lock) = new_lock {
-			locks.push(lock)
+			};
 		}
 
-		Self::update_locks(who, &locks);
+		generate_storage_types!("Instance0DarwiniaBalances", "TotalIssuance", OldRingTotalIssuance => Value<()>);
+		generate_storage_types!("Balances", "TotalIssuance", NewRingTotalIssuance => Value<()>);
+		generate_storage_types!("Instance1DarwiniaBalances", "TotalIssuance", OldKtonTotalIssuance => Value<()>);
+		generate_storage_types!("Kton", "TotalIssuance", NewKtonTotalIssuance => Value<()>);
```
AsceticBear, line 1597: No need to do the migration for `Account`, `Locks`?

```rs
/// The balance of an account.
///
/// NOTE: This is only used in the case that this pallet is used to store balances.
#[pallet::storage]
pub type Account<T: Config<I>, I: 'static = ()> =
	StorageMap<_, Blake2_128Concat, T::AccountId, T::BalanceInfo, ValueQuery>;

/// Any liquidity locks on some account balances.
/// NOTE: Should only be accessed when setting, changing and freeing a lock.
#[pallet::storage]
#[pallet::getter(fn locks)]
pub type Locks<T: Config<I>, I: 'static = ()> = StorageMap<
	_,
	Blake2_128Concat,
	T::AccountId,
	Vec<BalanceLock<T::Balance, T::BlockNumber>>,
	ValueQuery,
>;
```

AurevoirXavier: Just checking.

And I skip the map items checking.

AurevoirXavier:
P9 High

@AsceticBear @HackFisher

AsceticBear (approved):

HackFisher (approved):
LGTM.

Suggest do secondary check that module storage prefix names are the only storage migrations need to process.

PR #595, opened 2021-04-20
Test dvm migration use try-runtime

AsceticBear:
```sh
$ ./target/release/drml try-runtime live http://localhost:9933
2021-04-20 13:57:43 `pallet_timestamp::UnixTime::now` is called at genesis, invalid value returned: 0    
2021-04-20 13:57:43 [#0] 🗳  Finalized election round with compute ElectionCompute::OnChain.    
2021-04-20 13:57:43 [0] 💸 new validator set of size 1 has been processed for era 1    
2021-04-20 13:57:43 [darwinia-claims] Genesis Claims List is Set to EMPTY    
2021-04-20 13:57:43 initializing remote client to "http://localhost:9933"    
2021-04-20 13:57:43 scraping keypairs from remote node http://localhost:9933 @ 0x9de6f0e79a771f6fdbae5dadd58aef9f37abd8c225c9fcc072e5bf66a907f320    
2021-04-20 13:57:43 downloading data for all modules.    
2021-04-20 13:57:43 extending externalities with 1 manually injected keys    
2021-04-20 13:57:43 injecting a total of 692 keys    
2021-04-20 13:57:43 Schema migration successfully!    
2021-04-20 13:57:43 ⚠️ System declares internal migrations (which *might* execute), setting storage version to PalletVersion { major: 3, minor: 0, patch: 0 } 
```

### DIFF
--- a/frame/dvm/Cargo.toml
+++ b/frame/dvm/Cargo.toml
@@ -17,6 +17,7 @@ ethereum       = { version = "0.7.1", default-features = false, features = ["wit
 ethereum-types = { version = "0.11.0", default-features = false }
 evm            = { version = "0.25.0", default-features = false, features = ["with-codec"] }
 libsecp256k1   = { version = "0.3.5", default-features = false }
+log            = { version = "0.4.14", optional = true}
 rlp            = { version = "0.5.0", default-features = false }
 serde          = { version = "1.0.125", optional = true, default-features = false }
 sha3           = { version = "0.9.1", default-features = false }
@@ -76,4 +77,9 @@ substrate-std = [
 	"sp-io/std",
 	"sp-runtime/std",
 	"sp-std/std",
+]
+
+try-runtime = [
+	"log",
+	"frame-support/try-runtime",
 ]
--- a/frame/dvm/Cargo.toml
+++ b/frame/dvm/Cargo.toml
@@ -17,7 +17,7 @@ ethereum       = { version = "0.7.1", default-features = false, features = ["wit
 ethereum-types = { version = "0.11.0", default-features = false }
 evm            = { version = "0.25.0", default-features = false, features = ["with-codec"] }
 libsecp256k1   = { version = "0.3.5", default-features = false }
-log            = { version = "0.4.14", optional = true}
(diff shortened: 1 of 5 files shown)


frame/dvm/src/lib.rs
```diff
@@ -135,6 +134,7 @@ decl_storage! {
 			<Module<T>>::store_block(false);
 
 			// Initialize the storage schema at the well known key.
+			#[cfg(not(feature = "try-runtime"))]
```
AurevoirXavier, line 137: Why here

frame/dvm/Cargo.toml
```diff
@@ -17,6 +17,7 @@ ethereum       = { version = "0.7.1", default-features = false, features = ["wit
 ethereum-types = { version = "0.11.0", default-features = false }
 evm            = { version = "0.25.0", default-features = false, features = ["with-codec"] }
 libsecp256k1   = { version = "0.3.5", default-features = false }
+log            = { version = "0.4.14", optional = true}
```
AurevoirXavier, line 20: Lint

frame/dvm/src/lib.rs
AsceticBear: Since `PALLET_ETHEREUM_SCHEMA` is None in an existing network.  The test goal is to ensure this value from None to V1 after migration.

AurevoirXavier (approved):

AurevoirXavier:
Currently Darwinia is locked at tag v0.10.0

If this changes are necessary
A cherry-pick will need

AsceticBear:
> Currently Darwinia is locked at tag v0.10.0
> 
> If these changes are necessary
> A cherry-pick will need

What we do in this pr is just testing migration and code format. I think no need to do extra cherry-pick work.

AurevoirXavier:
> What we do in this pr is just testing migration and code format. I think no need to do extra cherry-pick work.

Don't you expect an migration testing on Crab?

HackFisher (approved):

AurevoirXavier:
I found a migration added. Not just testing.

Is this migration needed for Crab? @AsceticBear

PR #606, opened 2021-04-26
Capture runner call error

AsceticBear:
The runner call function always returns `Ok(...)` even an inner execution error happened.

We should capture the `exit_reason` contained in the call result and make a decision to commit or discard state change.


frame/evm/precompile/contracts/kton/Cargo.toml
```diff
@@ -16,6 +16,7 @@ codec          = { package = "parity-scale-codec", version = "2.0.1", default-fe
 ethereum-types = { version = "0.11.0", default-features = false }
 evm            = { version = "0.25.0", default-features = false, features = ["with-codec"] }
 hex            = { version = "0.4.3", default-features = false }
+log             = { version = "0.4.14" }
```
AurevoirXavier, line 19: lint

HackFisher (approved):

AurevoirXavier (approved):

PR #617, opened 2021-05-08
add msg.sender in ethereum backing lock event

xiaoch05:


frame/bridge/ethereum/issuing/src/lib.rs
```diff
@@ -236,27 +236,22 @@ impl<T: Config> Module<T> {
 
 	fn abi_encode_token_redeem(result: EthLog) -> Result<Vec<u8>, DispatchError> {
 		log::debug!("abi_encode_token_redeem");
-		let token_address = result.params[0]
-			.value
-			.clone()
-			.into_address()
-			.ok_or(<Error<T>>::AddressCF)?;
-		let dtoken_address = result.params[1]
+		let dtoken_address = result.params[2]
```
AsceticBear, line 239: It would be better to add some comments about what params represent? the `result.param[0]` is useless?

xiaoch05: ok I'll add some.
`result.param[0] `and `result.param[1]` is used by others. not here.

AsceticBear (approved):

HackFisher (approved):

HackFisher:
LGTM, but need to bump pangolin spec version in later PR before runtime upgrade to commits after this.

PR #624, opened 2021-05-12
Bridge Pangolin <-> Millau

AurevoirXavier:
#621
#624


bin/node/cli/src/service.rs
```diff
@@ -715,15 +741,26 @@ where
...
+			let timestamp = sp_timestamp::InherentDataProvider::from_system_time();
+			let slot =
+				sp_consensus_babe::inherents::InherentDataProvider::from_timestamp_and_duration(
+					*timestamp,
+					slot_duration,
+				);
+			let uncles =
+				sp_authorship::InherentDataProvider::<<Block as BlockT>::Header>::check_inherents();
+
+			Ok((timestamp, slot, uncles))
```
AsceticBear, line 762: Compared with https://github.com/paritytech/polkadot/tree/master/node#L1156 `Ok((timestamp, slot))`

The return value should not contain `uncles`?

AurevoirXavier: Because Polkadot doesn't require this `Inherents`

AurevoirXavier:
DO NOT merge.

bin/node/runtime/pangolin/src/pallets/bridge_dispatch.rs
```diff
@@ -0,0 +1,22 @@
+pub use pallet_bridge_dispatch::Instance1 as WithMillauDispatch;
```
fewensa, line 1:
```suggestion
pub use pallet_bridge_dispatch::Instance1 as WithPangolinDispatchInstance;
```

fewensa: It's right, but can change use same style? `WithMillauDispatchInstance` `WithMillauGrandpaInstance` `WithMillauMessagesInstance`

AurevoirXavier: I don't think so.

Checkout these:
https://github.com/darwinia-network/darwinia-common/blob/master/bin/node/runtime/pangolin/src/pallets/collective.rs#L2
https://github.com/paritytech/polkadot/blob/master/runtime/polkadot/src/lib.rs#L526

bin/node/runtime/pangolin/src/pallets/bridge_grandpa.rs
```diff
@@ -0,0 +1,27 @@
+pub use pallet_bridge_grandpa::Instance1 as MillauGrandpa;
```
fewensa, line 1:
```suggestion
pub use pallet_bridge_grandpa::Instance1 as WithMillauGrandpa;
```

same with bridge_dispatch (WithMillauDispatch) and bridge_messages (WithMillauGrandpa) or remove `With` in bridge_dispatch and bridge_messages

HackFisher (changes requested):
Need to bump pangolin spec version if we are going to do upgrade.

HackFisher (approved):

PR #625, opened 2021-05-18
remove pallet-shift-session-manager

fewensa:
we don't need pallet-shift-session-manager
related: https://github.com/paritytech/parity-bridges-common/issues/970


AsceticBear (approved):
LGTM

PR #628, opened 2021-05-19
Impl IssuingHandler for ()

AsceticBear:


frame/evm/src/lib.rs
```diff
@@ -148,6 +148,12 @@ pub trait IssuingHandler {
 	fn handle(address: H160, caller: H160, input: &[u8]) -> DispatchResult;
 }
 
+impl IssuingHandler for () {
+	fn handle(_: H160, _: H160, _: &[u8]) -> DispatchResult {
+		Ok(())
```
AurevoirXavier, line 153: Is this expected?

AsceticBear: `type IssuingHandler = ();` only be used to replace `EmptyIssuingHandler`, just for the test scenario.

AurevoirXavier: Please add some docs.

PR #629, opened 2021-05-19
Update DVM since 365

AsceticBear:
This pr contains a new pallet feature `dynamic-fee`.  validators can adjust evm gas price via adding `--target-gas-price xxx` when setup node. 

Gas price are adjusted for each block based on target_gas_price, which has been configured by the validator.


HackFisher (approved):
LGTM, but I'm not sure whether @AurevoirXavier is going to merge this after next Pangolin runtime #631

AsceticBear:
> LGTM, but I'm not sure whether @AurevoirXavier is going to merge this after next Pangolin runtime #631

Merge this firstly and upgrade to pangolin along with 631.

AurevoirXavier (approved):

PR #631, opened 2021-05-19
Substrate Update

AurevoirXavier:
https://github.com/darwinia-network/substrate-update-tracking/issues/17

PR #637, opened 2021-05-21
Expose SudoCall

fewensa:
millau -> pangolin init bridge need use it.


AurevoirXavier (changes requested):
Please revert the `Cargo.lock`

or 

PR to #631

PR #652, opened 2021-05-28
Abandon DarwiniaPrecompiles tuple

AsceticBear:
We can specify precompile contract address as any value after this pull request merged.


frame/evm/precompile/src/lib.rs
```diff
@@ -18,28 +18,40 @@
...
+			a if a == to_address(22) => Some(Kton::<R>::execute(input, target_gas, context)),
+			a if a == to_address(23) => Some(Issuing::<R>::execute(input, target_gas, context)),
+			_ => None,
+		}
+	}
+}
+
+fn to_address(a: u64) -> H160 {
+	H160::from_low_u64_be(a)
+}
```
HackFisher, line 57: I like this approach, but I think this code should located in pangolin runtime.

HackFisher (changes requested):

frame/evm/precompile/src/lib.rs
AsceticBear: Great catch. It would be much better to put this logic in runtime. Please review again.

HackFisher (approved):

HackFisher:
@AurevoirXavier @xiaoch05 please help review this.

xiaoch05 (approved):

bin/node/runtime/pangolin/src/pallets/evm.rs
```diff
@@ -1,11 +1,46 @@
...
+		input: &[u8],
+		target_gas: Option<u64>,
+		context: &Context,
+	) -> Option<core::result::Result<(ExitSucceed, Vec<u8>, u64), ExitError>> {
+		fn to_address(a: u64) -> H160 {
+			H160::from_low_u64_be(a)
+		}
+		match address {
+			// Ethereum precompiles
+			a if a == to_address(1) => Some(ECRecover::execute(input, target_gas, context)),
```
AurevoirXavier, line 30: `_ if address == to_address(1)` looks better

bin/node/runtime/pangolin/Cargo.toml
```diff
@@ -13,6 +13,7 @@ version    = "2.4.0"
 # crates
 array-bytes       = { version = "1.1.0" }
 codec             = { package = "parity-scale-codec", version = "2.1.1", default-features = false }
+evm-crate         = { default-features = false, version = "0.25.0", features = ["with-codec"], package = "evm" }
```
AurevoirXavier, line 16: Any reason rename evm to evm-crate?

AsceticBear: O, that's because the `crate evm` will conflict with the`runtime mod evm`.

AurevoirXavier (approved):

PR #654, opened 2021-06-01
Merge withdraw and kton into new transfer precompile

AsceticBear:
In this PR, I merged the original `withdraw` precompile(0x15) and the `kton` precompile(0x16) into a new precompile contract, named `transfer`, which address is 0x15, the same as with the original withdraw precompile address.  

Something important to notice:

I **did not change the protocol of old precompiles**. For application users who use kton precompile, just update the kton precompile address from 0x16 to 0x15. @hujw77  @sxlwar We should update wkton contract before this change goes to pangolin.


HackFisher:
One question, does `usize` have the same byte length in different platform? @AsceticBear @AurevoirXavier

AurevoirXavier:
> One question, does `usize` have the same byte length in different platforms? @AsceticBear @AurevoirXavier

https://matrix.to/#/!HzySYSaIhtyWrwiwEV:matrix.org/$162270534229674MqRiQ:matrix.org?via=matrix.parity.io&via=matrix.org&via=corepaper.org

AsceticBear:
> One question, does `usize` have the same byte length in different platforms? @AsceticBear @AurevoirXavier

Good questioin. 

The answer from AurevoirXavier Element:

```sh
> bear | Darwinia Network
Can i use usize type for number in pallet?
> Xavier - Darwinia.Network
Yes
But in WASM it's u32
So make sure it won't be grow up to u32::max_value + 1
```

I also find some example from substrate:

1. https://github.com/paritytech/substrate/blob/master/frame/authorship/src/lib.rs#L33
2. https://github.com/paritytech/substrate/blob/master/frame/elections/src/lib.rs#L153

In the end, I think this usage is ok.

bin/node/runtime/pangolin/src/pallets/evm_.rs
```diff
@@ -32,10 +31,7 @@ impl<R: dvm_ethereum::Config> PrecompileSet for PangolinPrecompiles<R> {
...
 			// Darwinia precompiles
 			_ if address == to_address(21) => {
-				Some(<WithDraw<R>>::execute(input, target_gas, context))
-			}
-			_ if address == to_address(22) => {
-				Some(<KtonPrecompile<R>>::execute(input, target_gas, context))
+				Some(<Transfer<R>>::execute(input, target_gas, context))
 			}
 			_ if address == to_address(23) => {
 				Some(<Issuing<R>>::execute(input, target_gas, context))
```
HackFisher, line 37: ```Issuing``` might need to be replaced by ```Dispatch``` Precompile, am I right?  @xiaoch05 @AsceticBear

xiaoch05: Yes, later it will be replaced.

frame/dvm/src/mock.rs
```diff
@@ -152,6 +154,17 @@ impl AddressMapping<AccountId32> for HashedAddressMapping {
...
+impl<T: darwinia_evm::Config> Precompile for RingBack<T> {
+	fn execute(
+		input: &[u8],
+		target_gas: Option<u64>,
+		context: &Context,
+	) -> core::result::Result<(ExitSucceed, Vec<u8>, u64), ExitError> {
+		darwinia_evm_precompile_transfer::ring::RingBack::<T>::transfer(&input, target_gas, context)
+	}
+}
+
```
HackFisher, line 167: Why can not use ```Transfer``` precompile for testing directly?

bin/node/runtime/pangolin/src/pallets/evm_.rs
AsceticBear: In my understanding, you are right! But there will be a `dispatch_wapper precompile` to replace the current `Issuing` contract.

xiaoch05: The Dispatch Precompile Wrapper coding will be completed with s2s backing & issuing PR, so in this pr, we keep it as the old version.

frame/evm/precompile/contracts/transfer/src/lib.rs
```diff
@@ -0,0 +1,71 @@
...
+use evm::{Context, ExitError, ExitSucceed};
+
+pub type AccountId<T> = <T as frame_system::Config>::AccountId;
+
+/// Transfer Precompile Contract, used to support the exchange of KTON and RING tranfer.
+pub enum Transfer<T> {
+	/// Transfer RING back from DVM to Darwinia
+	RingBack,
+	/// Transfer KTON between Darwinia and DVM contract
+	KtonTransfer,
```
HackFisher, line 42: Why not merge KTON action together here, it will look simpler.

This looks OK for me though.

frame/evm/precompile/contracts/transfer/src/lib.rs
```diff
@@ -0,0 +1,71 @@
...
+use ring::RingBack;
+// --- crate ---
+use evm::{Context, ExitError, ExitSucceed};
+
+pub type AccountId<T> = <T as frame_system::Config>::AccountId;
+
+/// Transfer Precompile Contract, used to support the exchange of KTON and RING tranfer.
+pub enum Transfer<T> {
+	/// Transfer RING back from DVM to Darwinia
+	RingBack,
```
HackFisher, line 40: Duplicate name with ```ring::RingBack```

HackFisher (changes requested):
@sxlwar please ACK these changes, might require companion PR in smart-app.

frame/evm/precompile/contracts/transfer/src/lib.rs
AsceticBear: Hmm, change to `RingTransfer` would be better. Let me fix it.

frame/evm/precompile/contracts/transfer/src/lib.rs
AsceticBear: I think it more clearly to separated the transfer by native token firstly, then individual actions for each token.

frame/dvm/src/mock.rs
AsceticBear: We have discussed a lot in the dev slack channel concerning this isssue.  Those comments might explain why I do it this way.

HackFisher (commented):
```
				<KtonRemainBalance as RemainBalanceOp<T, KtonBalance<T>>>::inc_remaining_balance(
					&wd.to_account_id,
					remain_balance.low_u128().saturated_into(),
				);
```
Why ring withdraw does not do this part?

AsceticBear:
> Why ring withdraw does not do this part?

Sadly, This is a bug that I am fixing it right now.

HackFisher:
> ```
> 				<KtonRemainBalance as RemainBalanceOp<T, KtonBalance<T>>>::inc_remaining_balance(
> 					&wd.to_account_id,
> 					remain_balance.low_u128().saturated_into(),
> 				);
> ```
> 
> Why ring withdraw does not do this part?

BTW, I think we need to consider the possibility of accumulating multiple remain_balances and causing this result larger than 1 1 * 10^9, in this case we should consider carry the overflow to Currency Balances, otherwise the balance will be locked in the remaining forever.

Current remaining balance is also using u64, which is too large for 10^9, can use u32 (max value 4294967295) is large enough for the storage.

HackFisher:
From @AurevoirXavier , for the reason of test compile failure: https://users.rust-lang.org/t/the-trait-xxx-is-not-implemented-for-xxx/60830/2?u=aurevoirxavier

AsceticBear:
> in this case we should consider carrying the overflow to Currency Balances

Actually, there are two cases:
1. For dvm account(H160), we have covered this case, the logic here https://github.com/darwinia-network/darwinia-common/blob/master/frame/dvm/src/account_basic.rs#L176
2. For accountId, we are not. An example here: https://github.com/darwinia-network/darwinia-common/blob/master/frame/evm/precompile/contracts/kton/src/lib.rs#L148, It can be fixed later, maybe next pull request.

> From @AurevoirXavier , for the reason of test compile failure: https://users.rust-lang.org/t/the-trait-xxx-is-not-implemented-for-xxx/60830/2?u=aurevoirxavier

Thanks so so so much! The lint will help me fix it.

HackFisher:
> > in this case we should consider carrying the overflow to Currency Balances
> 
> Actually, there are two cases:
> 
> 1. For dvm account(H160), we have covered this case, the logic here https://github.com/darwinia-network/darwinia-common/blob/master/frame/dvm/src/account_basic.rs#L176
> 2. For accountId, we are not. An example here: https://github.com/darwinia-network/darwinia-common/blob/master/frame/evm/precompile/contracts/kton/src/lib.rs#L148, It can be fixed later, maybe next pull request.

Since remaining balance is for both kinds of accounts, I would suggest to do some refactor for remaining balance related traits, to provide a more unified operation which only only dvm account can use, but also substrate account, this can also helps re-implementation of overflow carry logic.

I tend to think this is the fault in this PR's scope, and suggest to fix it in this PR, or at least before next runtime upgrade(can avoid possible migration efforts in future). Otherwise, please open and new issue to track it.

frame/dvm/Cargo.toml
```diff
@@ -39,10 +39,12 @@ sp-std           = { default-features = false, git = "https://github.com/darwini
 [dev-dependencies]
 # crates.io
 array-bytes = { version = "1.3.3" }
+ethabi      = { git = "https://github.com/darwinia-network/ethabi.git", branch = "xavier-no-std", default-features = false }
+hex         = "0.4.3"
```
AurevoirXavier, line 43: why hex?

AsceticBear: `hex` is used in file `tests.rs` -> `hex::encode(wkton_balance_input(sender.address)))`

AurevoirXavier: why not `array_bytes::bytes2hex`

AsceticBear: It probably works, let me have a try.

frame/evm/precompile/contracts/transfer/src/kton.rs
```diff
@@ -83,25 +65,20 @@ impl<T: Config + dvm_ethereum::Config> Precompile for Kton<T> {
...
-				)?;
+				T::KtonAccountBasic::transfer(&caller, &wkton, value)?;
 				// Call WKTON wrapped contract deposit
-				let precompile_address = H160::from_str(KTON_PRECOMPILE).unwrap_or_default();
-				let raw_input = make_call_data(context.caller, call_data.value)?;
+				let raw_input = make_call_data(caller, value)?;
 				if let Ok(call_res) = T::Runner::call(
-					precompile_address,
-					call_data.wkton_address,
+					array_bytes::hex_into_unchecked(TRANSFER_ADDR),
```
HackFisher, line 77: We need to be carefully when using `array_bytes::hex_into_unchecked` in runtime, since it is using `unwrap()` under the ground, which could break security audit rules.

But since `TRANSFER_ADDR` here is const and I think we are good. But reviewer should always be nervous on this usage.

HackFisher: I suggest to change to use `hex2bytes` here.

```
/// assert_eq!(
/// 	array_bytes::hex2bytes("0x4c6f7665204a616e6520466f7265766572").unwrap_or_default(),
/// 	*b"Love Jane Forever"
/// );
```

HackFisher (changes requested):

frame/evm/precompile/contracts/transfer/src/ring.rs
```diff
@@ -10,76 +10,60 @@
...
-		let context_value = context_value.low_u128().unique_saturated_into();
-
-		let result = T::RingCurrency::transfer(
-			&contract_address,
-			&input.dest,
-			context_value,
-			ExistenceRequirement::AllowDeath,
+		// Ensure the context address should be precompile address
+		ensure!(
+			source == array_bytes::hex_into_unchecked(TRANSFER_ADDR),
```
HackFisher, line 52: Same here.

AsceticBear: In this case, `array_bytes::hex_try_into(TRANSFER_ADDR).unwrap_or_default()` would be more appropriate?

HackFisher: Throw Error might be better

AsceticBear: Done!

HackFisher (approved):

AsceticBear:
@AurevoirXavier Please review this pr again.

script_tests/tests/1_test-transfer.js
```diff
@@ -141,7 +141,7 @@ describe("Test Transfer Balance", function () {
 
 	it("Get sender balance after withdraw", async function () {
 		const balanceFrom = web3.utils.fromWei(await web3.eth.getBalance(addressFrom), "ether");
-		expect(balanceFrom).to.be.equal("123366.78888448799999999");
+		expect(balanceFrom).to.be.equal("123366.78887448799999999");
```
AurevoirXavier, line 144: Is this expected?

AsceticBear: Yes, I change the transfer precompile ring transfer part execution fee from `10000` to `20000`, the same with kton transfer fee.

AurevoirXavier (approved):

PR #656, opened 2021-06-02
s2s issuing backing feature

xiaoch05:
Related https://github.com/darwinia-network/darwinia-launch/pull/2


frame/bridge/s2s/chain/src/lib.rs
```diff
@@ -0,0 +1,82 @@
...
+#[allow(non_camel_case_types)]
+pub enum PangolinSub2SubIssuingCall {
+	#[codec(index = 0)]
+	cross_receive((Token, EthereumAddress)),
+}
+
+pub type ChainSelector = [u8; 4];
+
+pub fn encode_relay_message<AccountId>(
+	selector: ChainSelector,
```
HackFisher, line 60: Do we really have to use chain selector here?

frame/bridge/s2s/chain/Cargo.toml
```diff
@@ -0,0 +1,37 @@
+[package]
+authors     = ["Darwinia Network <hello@darwinia.network>"]
+description = "FRAME pallet to s2s chain"
+edition     = "2018"
+homepage    = "https://darwinia.network/"
+license     = "GPL-3.0"
+name        = "darwinia-s2s-chain"
```
HackFisher, line 7: This is is not a substrate pallet?

Maybe we should move it to primitives if we need it.

frame/bridge/s2s/chain/src/lib.rs
xiaoch05: yes, we use it to tell us which chain to receive message. and this will be moved to runtime.

frame/bridge/s2s/issuing/src/lib.rs
```diff
@@ -0,0 +1,285 @@
...
+                log::info!("new s2s token has been registered, ingore response");
+            } else if &input[4..8] == burn_action {
+                //burn action
+                let burn_info = TokenBurnInfo::decode(&input[8..])
+                    .map_err(|_| Error::<T>::InvalidDecoding)?;
+                let recipient = Self::account_id_try_from_bytes(burn_info.recipient.as_slice())?;
+                let mut target: [u8;4] = Default::default();
+                target.copy_from_slice(&input[..4]);
+                Self::cross_send(target, burn_info.source, recipient, burn_info.amount)?;
+            }
```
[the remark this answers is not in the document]
xiaoch05: this comes from the `mapping-token-factory` contract. It only has these two situation.

frame/bridge/s2s/backing/src/lib.rs
```diff
@@ -0,0 +1,274 @@
...
+					decimal: 9,
+				}),
+			});
+
+			let message = (
+				target,
+				token.clone(),
+				RelayAccount::EthereumAccount(recipient),
+			);
+			T::IssuingRelay::relay_message(&message);
```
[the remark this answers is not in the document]
xiaoch05: This is not debugged. So i'm not sure the detail of the origin remote_chain received.
From the docs of the parity-bridge-common, the origin the remote_chain received is like this (origin+chainid).
Here, origin is Source Origin, and use the relay pallet as origin account.

frame/bridge/s2s/chain/src/lib.rs
HackFisher: Ah, I see, but fro each mapping token, we know its source chain right?

xiaoch05: yes. the target chain id is unique.

frame/bridge/s2s/chain/Cargo.toml
xiaoch05: yes it's not suitable to be here.
Should be moved to primitives and runtime.

HackFisher:
S2S_Backing_Issuing

frame/bridge/s2s/backing/src/lib.rs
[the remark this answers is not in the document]
HackFisher: I found an SourceRoot definition in https://github.com/paritytech/parity-bridges-common/blob/e74e8ab73f91eb8435dbe6f181d37cc2d5f34c70/primitives/message-dispatch/src/lib.rs#L71

Message in source chain is something like this:

```
let payload = send_message::message_payload(
			Default::default(),
			call.get_dispatch_info().weight,
			bp_message_dispatch::CallOrigin::SourceRoot,
			&call,
		);
```

https://github.com/paritytech/parity-bridges-common/blob/e74e8ab73f91eb8435dbe6f181d37cc2d5f34c70/modules/dispatch/src/lib.rs#L183

So if the source message is using Source Root, the dispatched target id will be
```
let hex_id = derive_account_id::<T::SourceChainAccountId>(source_chain, SourceAccount::Root);
				let target_id = T::AccountIdConverter::convert(hex_id);
```
https://github.com/paritytech/parity-bridges-common/blob/e74e8ab73f91eb8435dbe6f181d37cc2d5f34c70/primitives/runtime/src/lib.rs#L95

Or we can directly verify using ```pallet_bridge_dispatch::verify_message_origin```

frame/bridge/s2s/backing/src/lib.rs
```diff
@@ -0,0 +1,274 @@
...
+		#[pallet::weight(0)]
+		pub fn cross_receive(
+			origin: OriginFor<T>,
+			message: (Token, AccountId<T>),
+		) -> DispatchResultWithPostInfo {
+			let user = ensure_signed(origin)?;
+			// the s2s message relay has been verified the message comes from the issuing pallet with the
+			// chainID and issuing sender address.
+			// here only we need is to check the sender is in whitelist
+			let backing = T::IssuingRelay::verify(&user)?;
```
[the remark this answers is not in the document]
xiaoch05: here the origin is not Root. It is root + sourcechain, I think the whitelist is needed.

bin/node/runtime/pangolin/src/bridges/substrate/millau_messages.rs
```diff
@@ -247,3 +248,17 @@ impl SourceHeaderChain<bp_millau::Balance> for Millau {
 		)
 	}
 }
+
+pub struct MillauCallToPayload;
+
+impl CallToPayload<AccountId, ToMillauMessagePayload> for MillauCallToPayload {
+    fn to_payload(account: AccountId, call: Vec<u8>) -> ToMillauMessagePayload {
+        return FromThisChainMessagePayload::<WithMillauMessageBridge> {
+			spec_version: 1,
```
HackFisher, line 257: What's the meaning of the spec_version? Runtime spec version?

bin/node/runtime/pangolin/src/bridges/substrate/millau_messages.rs
```diff
@@ -247,3 +248,17 @@ impl SourceHeaderChain<bp_millau::Balance> for Millau {
...
+			weight: 100,
```
HackFisher, line 258: `spec_version` and `weight` should from Target chain(Crab DVM)'s Ethereum transaction input.

AsceticBear:
```
target chain mapping token manager list {

struct TokenInfo {
    uint8 token_type; // The type of the source token, e.g. ERC20 in DVM (0), or native token (ring/kton)(1) or parachain asssets(statemine) (2, )
    address token_index;    // The index of the source token, if type is ERC20, then it means the source contract address.
}

把各种各样的Assets规范(Nomalization)成ERC20，Adaptor


MessagePayload Spec

}
```

xiaoch05:
- [ ] use SourceRoot account to send relay message and delete the whitelist
- [ ] target chain calculate the converted origin and verify the SourceRoot
- [ ] in contract, save the token type. Native/Erc20/Parachain...
- [ ] use dispatch call instead of trait call for calling send_message
- [ ] check the spec_version's meaning and it cannot be hard code
- [ ] change the selector from bytes to hash such as `sha3::Keccak256::digest`
- [ ] use util precompile contract instead of dispatch wrapper precompile to encode the dispatch call
- [ ] move the remote call definition to runtime bind with relay pallet
- [ ] use terminology with technical paper as the name of the function
- [ ] delete the issuing handle precompile contract, also use dispatch precompile contract

xiaoch05:
there are some situations, the s2s relay message cannot be decoded or executed by the remote chain. Such as invalid spec_version, invalid dispatch call definition.

So do we need to add a nonce in App Message and save the Message in storage? and we need to relay another dispatch call to the remote chain to reply the result of the original message.

The end life of the App Message should be the time when it receives the remote reply.
If the error occurs on the remote chain, we should check the spec_version and remote call's definition. Upgrade runtime if necessary, and resend the Message by the same nonce.

relay

bin/node/runtime/pangolin/src/bridges/substrate/millau_messages.rs
```diff
@@ -247,3 +248,16 @@ impl SourceHeaderChain<bp_millau::Balance> for Millau {
...
 	}
 }
+
+pub struct MillauCallToPayload;
+
+impl CallToPayload<ToMillauMessagePayload> for MillauCallToPayload {
+	fn to_payload(spec_version: u32, call: Vec<u8>) -> ToMillauMessagePayload {
+		return FromThisChainMessagePayload::<WithMillauMessageBridge> {
+			spec_version,
+			weight: 100,
```
HackFisher, line 258: Is 100 enough for this `weight`?

 If it is less than actual static weight, the call is not dispatched. What's the actual static weight then?
```
/// Message payload type used by dispatch module.
#[derive(RuntimeDebug, Encode, Decode, Clone, PartialEq, Eq)]
pub struct MessagePayload<SourceChainAccountId, TargetChainAccountPublic, TargetChainSignature, Call> {
	/// Runtime specification version. We only dispatch messages that have the same
	/// runtime version. Otherwise we risk to misinterpret encoded calls.
	pub spec_version: SpecVersion,
	/// Weight of the call, declared by the message sender. If it is less than actual
	/// static weight, the call is not dispatched.
	pub weight: Weight,
	/// Call origin to be used during dispatch.
	pub origin: CallOrigin<SourceChainAccountId, TargetChainAccountPublic, TargetChainSignature>,
	/// The call itself.
	pub call: Call,
}
```

frame/bridge/s2s/backing/src/lib.rs
```diff
@@ -0,0 +1,252 @@
...
+
+	#[pallet::call]
+	impl<T: Config> Pallet<T> {
+		/// Lock token in this chain and cross transfer to the target chain
+		///
+		/// Target is the id of the target chain defined in s2s_chain pallet
+		// TODO: update the weight
+		#[pallet::weight(0)]
+		#[frame_support::transactional]
+		pub fn lock_and_cross_send(
```
HackFisher, line 146:
```suggestion
		pub fn lock_and_remote_issue(
```

frame/bridge/s2s/backing/src/lib.rs
```diff
@@ -0,0 +1,252 @@
...
+			);
+			T::IssuingRelay::relay_message(&message)?;
+			Self::deposit_event(Event::TokenLocked(token, user, recipient, amount));
+			Ok(().into())
+		}
+
+		/// Receive target chain locked message and unlock token in this chain.
+		// TODO: update the weight
+		#[pallet::weight(0)]
+		pub fn cross_receive_and_unlock(
```
HackFisher, line 197:
```suggestion
		pub fn remote_burn_and_unlock(
```

frame/bridge/s2s/issuing/src/lib.rs
```diff
@@ -0,0 +1,328 @@
...
+				return Err(<Error<T>>::InvalidAction.into());
+			}
+			Ok(().into())
+		}
+
+		/// Handle relay message sent from the source backing pallet with relay message
+		///
+		/// If the mapped token is not created, then create it.
+		#[pallet::weight(0)]
+		pub fn cross_receive_and_redeem(
```
HackFisher, line 115:
```suggestion
		pub fn cross_lock_and_issue(
```

It confuse me and says it is redeem  (burn and remote unlock)?

And there is a lazy token creation process here I can not clear say good

Suggest to split registration from cross_lock_and_issue first.

frame/bridge/s2s/issuing/src/lib.rs
```diff
@@ -0,0 +1,328 @@
...
+			symbol,
+			decimal,
+			backing,
+			address,
+		)
+		.map_err(|_| Error::<T>::InvalidEncodeERC20)?;
+		Ok(input)
+	}
+
+	fn encode_token_redeem(
```
HackFisher, line 273: rename

frame/bridge/s2s/relay/Cargo.toml
```diff
@@ -0,0 +1,65 @@
+[package]
+authors     = ["Darwinia Network <hello@darwinia.network>"]
+description = "FRAME pallet to s2s relay"
+edition     = "2018"
+homepage    = "https://darwinia.network/"
+license     = "GPL-3.0"
+name        = "darwinia-s2s-relay"
```
HackFisher, line 7: https://github.com/darwinia-network/parity-bridges-common/blob/master/modules/grandpa/src/lib.rs

I don't this this pallet is like relay, it is more like to be dispatch:

https://github.com/darwinia-network/parity-bridges-common/blob/master/modules/dispatch/src/lib.rs

Please fork Dispatch pallet or use Dispatch directly. There are all internal calls which can stay at some pallet else?

xiaoch05: this pallet do two things.
dispatch the relay-message and verify the message from remote chain.
So these two functions is more like a relay.

frame/bridge/s2s/issuing/src/lib.rs
xiaoch05: The registration will be split from remote_lock_and_issue. I also can not clearly say it better or not. Because it's the native token, and only be registered once.

frame/bridge/s2s/relay/Cargo.toml
HackFisher: Send message should use `pallet-bridge-messages`, right?


```
#[weight = T::WeightInfo::send_message_weight(payload)]		pub fn send_message(			origin,			lane_id: LaneId,			payload: T::OutboundPayload,			delivery_and_dispatch_fee: T::OutboundMessageFee,		) -&gt; DispatchResult {
```

https://github.com/darwinia-network/parity-bridges-common/blob/b9545dcb66da4ea572c9c43959cf088f2047a2c5/modules/messages/src/lib.rs#L296

frame/bridge/s2s/relay/src/lib.rs
```diff
@@ -0,0 +1,179 @@
...
+	fn relay_message(message: &Self::RelayMessage) -> Result<(), DispatchError> {
+		let (spec_version, token, relay_account) = message.clone();
+		let encoded = T::BridgedAssetReceiverT::encode_call(token, relay_account)
+			.map_err(|_| <Error<T, I>>::EncodeInvalid)?;
+		let relay_id: AccountId<T> = T::PalletId::get().into_account();
+		let payload = T::CallToPayload::to_payload(spec_version, encoded);
+
+		T::MessageRelayCallT::encode_call(payload)
+			.dispatch(RawOrigin::Signed(relay_id).into())
+			.map_err(|_| <Error<T, I>>::DispatchFailed)?;
```
HackFisher, line 153: Is here going to do remote call or local call?

If it is an remote call, why there is no pallet-bridge-messages's `send_message `?

Dispatch and Message are two different things, Dispatch is to call send from remote, Message is it to send call as message to remote.

This is why I think we do not need this pallet, we should use `pallet-bridge-messages` and `pallet-bridge-dispatch` directly from Backing and Issuing

HackFisher:
Here is an diagram explain the pallets and its use cases:

s2s_remote_chain_call

frame/bridge/s2s/relay/src/lib.rs
xiaoch05: It is local call. Yes there is no trait for pallet-bridge-messages's `send_message`. And we don't want to modify pallet-bridge-messages's code. So we use a dispatch call here.

xiaoch05:
Some TODO things
- [ ] check when revert occurs in the contract, what will happen in pallet. @AsceticBear 
- [ ] optimization for backing pallet's two encode trait. merge them to a single trait. @AsceticBear 
- [ ] the dvm will execute precompile many times to estimate gas. We should check whether it'll send the relay message many times. @xiaoch05 
- [ ] the security of dispatch precompile and contract's interactive. @AsceticBear @xiaoch05 
- [ ] how to set the parameters `weight` and `fee`. @xiaoch05 @HackFisher 
- [ ] Need to give an address for the Root account. And transfer some fee to it. @xiaoch05

xiaoch05:
* the fee is determined by the user when sending relay message, https://github.com/darwinia-network/parity-bridges-common/blob/master/modules/messages/src/lib.rs#L300
* the fee will be transferred to the `relayer_fund_account` https://github.com/darwinia-network/parity-bridges-common/blob/master/modules/messages/src/instant_payments.rs#L82, and distributed as rewards to relayers when the relay message confirmed. https://github.com/darwinia-network/parity-bridges-common/blob/master/modules/messages/src/instant_payments.rs#L141
* it has a minimal dispatch fee acquire related to the weight of the remote dispatch call and the message size. https://github.com/darwinia-network/parity-bridges-common/blob/master/bin/runtime-common/src/messages.rs#L277
* fee estimate
  * transaction_fee = basefee(125000000) + len_fee(1000~2000) + weight_fee(multiplier * (delivery_weight + remote_call_weight))
  * total_fee = delivery_transaction + confirmation_transaction
  * here we set multiplier = 0, the minimal_fee estimate is about 250,002,000~ 250,004,000

bin/node/runtime/pangolin/src/pallets/s2s_issuing.rs
```diff
@@ -11,16 +11,13 @@ use darwinia_support::s2s::{RelayMessageCaller, TruncateToEthAddress};
...
+		fee: Balance,
 	) -> Result<PostDispatchInfo, DispatchErrorWithPostInfo<PostDispatchInfo>> {
-		let call: Call = BridgeMessagesCall::<Runtime, Millau>::send_message(
-			[0; 4],
-			payload,
-			300_000_000u128.into(),
-		)
-		.into();
+		let call: Call =
+			BridgeMessagesCall::<Runtime, Millau>::send_message([0; 4], payload, fee).into();
```
AsceticBear, line 20: `[0;4]` lane id should not be hardcoded here.

AsceticBear:
TODO LIST:

- [ ] weight issue

frame/bridge/s2s/issuing/src/lib.rs
```diff
@@ -0,0 +1,374 @@
...
+	#[pallet::call]
+	impl<T: Config> Pallet<T> {
+		/// Handle dispatch call from dispatch precompile contract
+		///
+		/// When user burn their tokens, this handler will receive the event from dispatch
+		/// precompile contract, and relay this event to the target chain to unlock asset.
+		// TODO: update the weight
+		#[pallet::weight(0)]
+		#[frame_support::transactional]
+		pub fn dispatch_handle(origin: OriginFor<T>, input: Vec<u8>) -> DispatchResultWithPostInfo {
```
HackFisher, line 103: Request better document, function name and weight.

The reason we need this `dispatch_handle` is that the parameters passed from precompile is `input: Vec<u8>`, and we need a handle to parse it?

frame/bridge/s2s/issuing/src/lib.rs
```diff
@@ -0,0 +1,374 @@
...
+				Self::burn_and_remote_unlock(fee, burn_info)?;
+			} else {
+				log::trace!("No action match this input selector");
+			}
+			Ok(().into())
+		}
+
+		/// Handle remote register relay message
+		/// Before the token transfer, token should be created first
+		#[pallet::weight(0)]
```
HackFisher, line 140: Checking weight?

frame/bridge/s2s/issuing/src/lib.rs
```diff
@@ -0,0 +1,374 @@
...
+		pub fn remote_register(origin: OriginFor<T>, token: Token) -> DispatchResultWithPostInfo {
```
HackFisher, line 141:
```suggestion
		pub fn registerf_from_remote(origin: OriginFor<T>, token: Token) -> DispatchResultWithPostInfo {
```

frame/bridge/s2s/issuing/src/lib.rs
```diff
@@ -0,0 +1,374 @@
...
+			let user = ensure_signed(origin)?;
+			let backing = Self::verify_origin(&user)?;
```
HackFisher, line 143: backing or SourceRoot? Do we need to verify it is SourceRoot here?

frame/bridge/s2s/issuing/src/lib.rs
```diff
@@ -0,0 +1,374 @@
...
+					));
+				}
+				_ => return Err(Error::<T>::InvalidTokenOption.into()),
+			}
+			Ok(().into())
+		}
+
+		/// Handle relay message sent from the source backing pallet with relay message
+		#[pallet::weight(0)]
+		pub fn remote_issue(
```
HackFisher, line 183:
```suggestion
		pub fn issue_from_remote(
```

frame/bridge/s2s/issuing/src/lib.rs
```diff
@@ -0,0 +1,374 @@
...
+			token_type,
+			address,
+			amount,
+			account_id.into(),
+			burn_info.fee,
+		));
+		Ok(())
+	}
+
+	fn verify_origin(account: &T::AccountId) -> Result<H160, DispatchError> {
```
HackFisher, line 358:
```suggestion
	fn ensure_source_root(account: &T::AccountId) -> Result<H160, DispatchError> {
```

frame/bridge/s2s/issuing/src/lib.rs
HackFisher: It will clear if we change the name to `ensure_source_root `.

```suggestion
			Self::ensure_source_root(&user)?;
```

xiaoch05: It will verify the SourceRoot identity and return an ethereum address called backing address derived from the SourceRoot address and the remote chainid. If the same source ERC20 token address comes from different remote chains, we use the backing address to generate different mapping token addresses.

HackFisher:
命名上还想再优化和讨论一下，下面两种方式，看看哪种比较好，我更倾向于B，更加准确和富信息一些。

A:
```
// Backing
pub fn lock_and_remote_issue(...)

pub fn unlock_from_remote(...)


// Issuing
pub fn issue_from_remote(...)

pub fn burn_and_remote_unlock(...)
```

B:
```
// Backing
pub fn lock_and_target_issue(...)

pub fn unlock_from_source(...)


// Issuing
pub fn issue_from_source(...)

pub fn burn_and_target_unlock(...)
```

AsceticBear:
I think B is better.

xiaoch05:
> 命名上还想再优化和讨论一下，下面两种方式，看看哪种比较好，我更倾向于B，更加准确和富信息一些。
> 
> A:
> 
> ```
> // Backing
> pub fn lock_and_remote_issue(...)
> 
> pub fn unlock_from_remote(...)
> 
> 
> // Issuing
> pub fn issue_from_remote(...)
> 
> pub fn burn_and_remote_unlock(...)
> ```
> 
> B:
> 
> ```
> // Backing
> pub fn lock_and_target_issue(...)
> 
> pub fn unlock_from_source(...)
> 
> 
> // Issuing
> pub fn issue_from_source(...)
> 
> pub fn burn_and_target_unlock(...)
> ```

我比较倾向 A
source 和 target 也可能是指 local

frame/bridge/s2s/issuing/src/lib.rs
xiaoch05: this function is called by dispatch precompile contract to receive the burn event of the mapped token and parse it.
how about changing it to `asset_burn_event_handle`?

HackFisher: `asset_burn_event_handle` sound good.

frame/bridge/s2s/issuing/src/lib.rs
HackFisher: OK

HackFisher:
> 命名上还想再优化和讨论一下，下面两种方式，看看哪种比较好，我更倾向于B，更加准确和富信息一些。
> 
> A:
> 
> ```
> // Backing
> pub fn lock_and_remote_issue(...)
> 
> pub fn unlock_from_remote(...)
> 
> 
> // Issuing
> pub fn issue_from_remote(...)
> 
> pub fn burn_and_remote_unlock(...)
> ```
> 
> B:
> 
> ```
> // Backing
> pub fn lock_and_target_issue(...)
> 
> pub fn unlock_from_source(...)
> 
> 
> // Issuing
> pub fn issue_from_source(...)
> 
> pub fn burn_and_target_unlock(...)
> ```

最后决定用A.

AsceticBear:
**Something Note For Reviewers And Developers: Check `type.json`.**

xiaoch05:
we merge the branch to s2s branch first.
https://github.com/darwinia-network/darwinia-common/pull/691

AsceticBear:
This one can be closed now?

PR #662, opened 2021-06-07
Update `array-bytes`

AurevoirXavier:


bin/node/cli/src/chain_spec.rs
```diff
@@ -181,8 +181,8 @@ fn pangolin_build_spec_genesis() -> pangolin_runtime::GenesisConfig {
 	}
 	impl Keys {
 		fn new(sr25519: &str, ed25519: &str) -> Self {
-			let sr25519 = array_bytes::hex2array_unchecked!(sr25519, 32);
-			let ed25519 = array_bytes::hex2array_unchecked!(ed25519, 32);
+			let sr25519 = array_bytes::hex2array_unchecked(sr25519);
```
HackFisher, line 184:
```
pub fn hex2array_unchecked<H, const N: usize>(hex: H) -> [u8; N]
where
	H: AsRef<str>,
{
	hex2bytes_unchecked(hex).try_into().unwrap()
}
```

I'm curious how compiler can infer `N` from `&str` whose size is unknown in compile time as I understand.

AurevoirXavier: Depend on the later usage (require 32).

Also `unchecked` means it won't check the `&str`'s length and always try to cast it to 32, may cause panic.

frame/claims/src/lib.rs
```diff
@@ -990,23 +988,23 @@ mod tests {
...
 
 	#[test]
 	fn real_tron_sig_works() {
 		new_test_ext().execute_with(|| {
 			// "Pay RUSTs to the TEST account:0c0529c66a44e1861e5e1502b4a87009f23c792518a7a2091363f5a0e38abd57"
-			let sig = array_bytes::hex2array_unchecked!("0x34c3d5afc7f8fa08f9d00a1ec4ac274c63ebce99460b556de85258c94f41ab2f52ad5188bd9fc51251cf5dcdd53751b1bd577828db3f2e8fe8ef77907d7f3f6a1b", 65);
+			let sig = array_bytes::hex2array_unchecked("0x34c3d5afc7f8fa08f9d00a1ec4ac274c63ebce99460b556de85258c94f41ab2f52ad5188bd9fc51251cf5dcdd53751b1bd577828db3f2e8fe8ef77907d7f3f6a1b");
 			let sig = EcdsaSignature(sig);
-			let who = array_bytes::hex2array_unchecked!("0x0c0529c66a44e1861e5e1502b4a87009f23c792518a7a2091363f5a0e38abd57", 32).using_encoded(to_ascii_hex);
+			let who = array_bytes::hex2array_unchecked::<_, 32>("0x0c0529c66a44e1861e5e1502b4a87009f23c792518a7a2091363f5a0e38abd57").using_encoded(to_ascii_hex);
```
AsceticBear, line 1005: Sometimes, we still need to add length param?

AurevoirXavier: Sometimes it can not deduce. Just like `Vec<_>`

PR #663, opened 2021-06-07
Replace all hardcode selectors with calculator

AsceticBear:
This pr will update after the main `s2s-issuing-backing branch` does more update later.


AsceticBear:
@xiaoch05 This is ready for review and merging.

PR #697, opened 2021-07-07
Correct Collective Genesis

AurevoirXavier:
We should initialize the collective members from these 2 pallets
1. Membership [Genesis](https://github.com/paritytech/substrate/blob/master/frame/membership/src/lib.rs#L88) <-> [Runtime Configuration](https://github.com/darwinia-network/darwinia-common/blob/master/bin/node/runtime/pangolin/src/pallets/membership.rs#L26)
2. Elections phragmen [Genesis](https://github.com/paritytech/substrate/blob/master/frame/elections-phragmen/src/lib.rs#L699) <-> [Runtime Configuration](https://github.com/darwinia-network/darwinia-common/blob/master/bin/node/runtime/pangolin/src/pallets/elections_phragmen.rs#L27)

### DIFF
--- a/bin/node/cli/src/chain_spec.rs
+++ b/bin/node/cli/src/chain_spec.rs
@@ -511,16 +511,19 @@ fn pangolin_development_genesis() -> pangolin_runtime::GenesisConfig {
 		pallet_im_online: Default::default(),
 		pallet_authority_discovery: Default::default(),
 		darwinia_democracy: Default::default(),
-		pallet_collective_Instance1: pangolin_runtime::CouncilConfig {
-			phantom: PhantomData::<pangolin_runtime::CouncilCollective>,
+		pallet_collective_Instance1: Default::default(),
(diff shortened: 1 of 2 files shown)


bin/node/cli/src/chain_spec.rs
```diff
@@ -511,16 +511,19 @@ fn pangolin_development_genesis() -> pangolin_runtime::GenesisConfig {
 		pallet_im_online: Default::default(),
 		pallet_authority_discovery: Default::default(),
 		darwinia_democracy: Default::default(),
-		pallet_collective_Instance1: pangolin_runtime::CouncilConfig {
-			phantom: PhantomData::<pangolin_runtime::CouncilCollective>,
-			members: collective_members.clone(),
```
AsceticBear, line 516: Do not need this genesis config now? Give more information to help with the review.

PR #703, opened 2021-07-09
Update dvm from 399

AsceticBear:
This pr is mainly used for synchronizing changes from 399 to 410 in the upstream library.


HackFisher (approved):
LGTM, but require more decription and campanion frontier PR.

bin/node/runtime/pangolin/src/pallets/system/mod.rs
```diff
@@ -9,7 +9,7 @@ use sp_version::RuntimeVersion;
 use crate::*;
 
 frame_support::parameter_types! {
-	pub const BlockHashCount: BlockNumber = 2400;
+	pub const BlockHashCount: BlockNumber = 256;
```
AsceticBear, line 12: Here is more information about this change: https://github.com/paritytech/frontier/pull/409

frame/evm/precompile/contracts/blake2/src/lib.rs
```diff
@@ -39,7 +38,7 @@ impl Precompile for Blake2F {
 		input: &[u8],
 		target_gas: Option<u64>,
 		_context: &Context,
-	) -> core::result::Result<(ExitSucceed, Vec<u8>, u64), ExitError> {
+	) -> core::result::Result<PrecompileOutput, ExitError> {
```
xiaoch05, line 41: how to use this Result in contract. This change will cause some issues with the old usage of precompile contract.

AsceticBear: https://github.com/rust-blockchain/evm/blob/master/src/executor/stack/mod.rs#L84-L89
```rs
pub struct PrecompileOutput {
	pub exit_status: ExitSucceed,
	pub cost: u64,
	pub output: Vec<u8>,
	pub logs: Vec<Log>,
}
```

The PrecompileOutput contains all fields in the old version, only add `log` field.

bin/node/runtime/pangolin/src/pallets/system/mod.rs
AsceticBear: todo: create an issue to trace this change in darwinia and crab.

AsceticBear: Trace issue: https://github.com/darwinia-network/darwinia/issues/680

frame/evm/precompile/contracts/blake2/src/lib.rs
xiaoch05: there is no change for the return value of the Precompile Contract.  So we need nothing to be changed. 
The struct PrecompileOutput will be parsed in the evm inner method. And only the two parameters `exit_status` and `output` will be transparent to our contract, it's the same as the older version.
https://github.com/rust-blockchain/evm/pull/32/files#diff-8b611944560f538201b7edeb4c936bf728d167100a3c97b80f6ba758ac0eb240R534-R552

frame/dvm/src/lib.rs
```diff
@@ -107,13 +106,24 @@ pub mod pallet {
...
 	impl<T: Config> Hooks<BlockNumberFor<T>> for Pallet<T> {
-		fn on_finalize(_block_number: T::BlockNumber) {
+		fn on_finalize(n: T::BlockNumber) {
 			<Pallet<T>>::store_block(
 				dp_consensus::find_pre_log(&<frame_system::Pallet<T>>::digest()).is_err(),
 				U256::from(UniqueSaturatedInto::<u128>::unique_saturated_into(
 					frame_system::Pallet::<T>::block_number(),
 				)),
 			);
+			// move block hash pruning window by one block
```
AurevoirXavier, line 116: https://github.com/paritytech/substrate/blob/master/frame/system/src/lib.rs#L1365-L1372

AurevoirXavier (approved):

Issue #702, opened 2021-07-08
Ethereum London Hardfork support


HackFisher:
https://blog.ethereum.org/2021/06/18/london-testnets-announcement/

# Include:
EIP-1559: Fee market change for ETH 1.0 chain
EIP-3198: BASEFEE opcode
EIP-3529: Reduction in refunds
EIP-3541: Reject new contracts starting with the 0xEF byte
EIP-3554: Difficulty Bomb Delay to December 1st 2021

# Affected(Current found)
- [ ] https://eips.ethereum.org/EIPS/eip-1559 #701 
- [ ] https://eips.ethereum.org/EIPS/eip-3554

References:

1. https://github.com/openethereum/openethereum/blob/fdaee51ca07710081d27722449478b20dfbf72c9/crates/ethcore/types/src/header.rs#L374-L376
2. https://github.com/openethereum/openethereum/blob/fdaee51ca07710081d27722449478b20dfbf72c9/crates/ethcore/res/chainspec/ropsten.json#L21
3. https://github.com/openethereum/openethereum/blob/193b25a22d5ff07759c6431129e95235510516f9/crates/ethcore/src/ethereum/ethash.rs#L80

HackFisher:
https://github.com/darwinia-network/darwinia-common/blob/d8896a3621cd8d6f00fc34238403720b62aad9c1/primitives/ethereum-primitives/src/pow.rs#L130

Issue #701, opened 2021-07-08
Support EIP-1559


AurevoirXavier:
https://eips.ethereum.org/EIPS/eip-1559
https://github.com/openethereum/openethereum/pull/393

Issue #271, opened 2020-09-07
Remove EthereumHeader::from_str_unchecked


HackFisher:

AurevoirXavier:
#117

PR #708, opened 2021-07-12
Refactor Ethereum Primitives

AurevoirXavier:
- [x] Drop duplicate primitive library.
	```rust
	// e.g.
	// `H64`, `H256` from different library
	use primitive_types::H64;
	use ethereum_types::H256;
	```
- [x] Clean unnecessary dependencies speed up the compiling.
- [x] Generic library with multi-features control, easy to use for side projects such as bridger, shadow, etc.
- [x] Better error handling.
- [x] #271
  - [x] New serializer/deserializer accept all Infura API like header spec. With this feature added, shadow does not need to handle the [`DarwiniaHeader` spec](https://sourcegraph.com/github.com/darwinia-network/shadow/-/blob/ffi/pkg/shadow/eth/header.go?L65-99). And we could finish the serialize/deserialize in bridger part with library `ethereum-primitives`. cc @xiaoch05 
- [x] #702 
  - [x] #701
    - [x] On-chain storage checks. ~~There is a new field named `base_fee_per_gas`, which type is an `Option`. Maybe it's compatible with scale-codec.~~
    - [x] Migration to new header spec.
      - [x] New header spec.
      - [x] Impl RLP encoder/decoder for new header.
      - [x] Reset `Relay::ConfirmedHeaderParcels`.
      - [x] Reset `RelayerGame::Affirmations`.
  - [x] Support EIP-3554.
    - [x] Add difficulty bomb delay.
- [ ] Maybe #655.


HackFisher:
Ethereum and BSC may use different version of primitives, not reasonable to merge and unify them, any substantial upgrade from chain side could break this.

AurevoirXavier:
```rust
panicked at 'called `Result::unwrap()` on an `Err` value: Error { cause: Some(Error { cause: Some(Error { cause: None, desc: "Not enough data to fill buffer" }), desc: "Could not decode variant byte for `Option`" }), desc: "Could not decode `Header::hash`" }'
```
EIP-1559 introduces a new field in the header `base_fee_per_gas`, which will break the scale-type encoded storage.

For us, we could clear the `ConfirmedHeaderParcels` storage then apply the new type. Or define a new storage name. (e.g. `EIP1559Header`).

@HackFisher

HackFisher:
> ```rust
> 
> panicked at 'called `Result::unwrap()` on an `Err` value: Error { cause: Some(Error { cause: Some(Error { cause: None, desc: "Not enough data to fill buffer" }), desc: "Could not decode variant byte for `Option`" }), desc: "Could not decode `Header::hash`" }'
> 
> ```
> 
> EIP-1559 introduces a new field in the header `base_fee_per_gas`, which will break the scale-type encoded storage.
> 
> 
> 
> For us, we could clear the `ConfirmedHeaderParcels` storage then apply the new type. Or define a new storage name. (e.g. `EIP1559HeaderParcels`).
> 
> 
> 
> cc @HackFisher 

Prefer clear & migrate the ConfirmedHeaderParcels storage then apply the new type.

AurevoirXavier:
The genesis's MMR root is wrong.
Need shadow to provide the right one.

cc @xiaoch05

primitives/merkle-patricia-trie/src/error.rs
```diff
@@ -22,35 +22,26 @@ extern crate alloc;
...
 	fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-		let printable = match *self {
-			TrieError::DB(ref err) => format!("trie error: {:?}", err),
-			TrieError::Decoder(ref err) => format!("trie error: {:?}", err),
-			TrieError::InvalidData => "trie error: invalid data".to_owned(),
-			TrieError::InvalidStateRoot => "trie error: invalid state root".to_owned(),
-			TrieError::InvalidProof => "trie error: invalid proof".to_owned(),
-		};
-		write!(f, "{}", printable)
+		fmt::Debug::fmt(&self, f)
```
AsceticBear, line 40: In this way, the output of `TrieError` will be the same with `#[derive(Debug)]`.How about removing `impl fmt::Display for TrieError {...}`?

AurevoirXavier: Out of scope of this PR.

There is an incoming refactor about `merkle-patricia-trie`.

frame/bridge/ethereum/linear-relay/src/lib.rs
```diff
@@ -106,7 +106,7 @@ pub trait Config: frame_system::Config {
 
 	type Event: From<Event<Self>> + Into<<Self as frame_system::Config>::Event>;
 
-	type EthereumNetwork: Get<EthereumNetworkType>;
+	type TargetNetwork: Get<EthereumNetwork>;
```
HackFisher, line 109:
```suggestion
	type BridgedNetwork: Get<EthereumNetwork>;
```

`Target` might be confused with target chain, this ethereum network here is actually source chain.

frame/bridge/ethereum/backing/src/lib.rs
```diff
@@ -28,8 +28,8 @@ pub use weights::WeightInfo;
 
 #[cfg(test)]
 mod mock;
-#[cfg(test)]
-mod test_with_linear_relay;
+// #[cfg(test)]
+// mod test_with_linear_relay;
```
HackFisher, line 32: linear relay pallet can be deleted.

AurevoirXavier: Would like to do some code cleaning in another PR.

HackFisher (approved):

bin/node/runtime/pangolin/src/lib.rs
```diff
@@ -898,6 +898,71 @@ impl dvm_rpc_runtime_api::ConvertTransaction<OpaqueExtrinsic> for TransactionCon
 	}
 }
 
+const PARCEL: &[u8] = &[
```
HackFisher, line 901: Put this inside test function if this is only used for test?

AurevoirXavier: Not only for the test but also [here](https://github.com/darwinia-network/darwinia-common/pull/708/files/a87307ede483ea3ac52a2f5d6545265cc5e92791#diff-d76e6add241f9f0be48db383dc3c10d6cd079d48703a6972e3ef9f3e64e29d80R984)

frame/bridge/relayer-game/src/lib.rs
```diff
@@ -20,6 +20,30 @@
...
+pub mod migration {
+	// --- darwinia-network ---
+	use crate::*;
+
+	pub fn migrate<T, I>()
+	where
+		T: Config<I>,
+		I: Instance,
+	{
+		<RelayHeaderParcelToResolve<T, I>>::put(<Vec<RelayHeaderId<T, I>>>::new());
```
AsceticBear, line 32: why not `remove_all()` here?

AurevoirXavier: This is a value, not a map. 

I think you mean: `<RelayHeaderParcelToResolve<T, I>>::kill()`.

But it's okay to put an empty `Vec` into it.

bin/node/runtime/pangolin/src/lib.rs
AsceticBear: How about put the  `PARCEL` into `darwinia_ethereum_relay::migration()`? It looks a bit wired to see this test here.

```rs
darwinia_ethereum_relay::migration::migrate() {
        let PARCEL = ...
}
```

bin/node/runtime/pangolin/src/lib.rs
```diff
@@ -921,10 +981,8 @@ impl OnRuntimeUpgrade for CustomOnRuntimeUpgrade {
 		// let _ = Ring::deposit_creating(&BridgeMillauMessages::relayer_fund_account_id(), 1 << 50);
 		// --->
 
-		darwinia_header_mmr::migration::initialize_new_mmr_state::<Runtime>(b"HeaderMMR", 50);
-		darwinia_relay_authorities::migration::migrate::<Runtime, EthereumRelayAuthoritiesInstance>(
-			b"Instance1DarwiniaRelayAuthorities",
-		);
+		darwinia_ethereum_relay::migration::migrate(PARCEL);
+		darwinia_relayer_game::migration::migrate::<Runtime, EthereumRelayerGameInstance>();
```
AsceticBear, line 985: Can we remove the commented line `<-- Hack for local test ... --->` in `on_runtime_upgrade()` and `pre_upgrade()` now?

AurevoirXavier: I think we can't. Until the parity-bridges-common fix that. Have you tried that?

Nodes will panic without those codes.

AsceticBear (approved):

AurevoirXavier:
It's safe to use `remove_all` and `kill`.

https://crates.parity.io/sp_externalities/trait.Externalities.html#tymethod.clear_prefix
https://crates.parity.io/frame_support/storage/trait.StorageValue.html#impl-StorageValue%3CT%3E

primitives/ethereum/src/pow.rs
```diff
@@ -105,6 +110,7 @@ impl EthashPartial {
 				m.insert(4370000, 3000000);
 				m.insert(7280000, 2000000);
 				m.insert(0x8c6180, 0x3d0900);
+				m.insert(0xc3d0e8, 0xaae60);
```
HackFisher, line 113: This looks not right, compared to https://github.com/openethereum/openethereum/pull/561/files#diff-06bc3942f66cb315836147192afb41985b91ee270161b398f69a2d07fd4ddc05R141

AurevoirXavier: I forgot where I got these values.
Should we update this to the latest openethereum code in the next runtime upgrade?

HackFisher: Yes, I double checked with geth, openethereum's code should be right.

AurevoirXavier: Should this be included in darwinia-v0.11.7?
I'll fire a PR right now.

HackFisher: OK, darwinia-v0.11.7 branch is a good place, and don't forget we also need it in master.

PR #711, opened 2021-07-14
Tag Thirdparty Libraries

AurevoirXavier:
Anchor the libraries to a particular tag. Avoid getting updated accidentally while running `cargo update` globally.


HackFisher (commented):
LGTM

PR #713, opened 2021-07-15
Update dvm transact and call usage

AsceticBear:
- Internal transaction vs  RPC transaction

The difference between internal transaction and RPC transaction is that internal transaction is able to catch evm runner execution exception and transfer to dispatch error, the RPC transaction can not.

- raw_call

An internal call is used to make a static call to a contract, whose sender is always `INTERNAL_CALLER`.


frame/dvm/src/lib.rs
```diff
@@ -492,12 +385,43 @@ impl<T: Config> Pallet<T> {
...
+			ExitReason::Succeed(_) => Ok(Some(T::GasWeightMapping::gas_to_weight(
+				used_gas.unique_saturated_into(),
+			))
+			.into()),
+			_ => Err(Error::<T>::InvalidCall.into()),
+		}
+	}
+
+	/// Pure read-only call to contract
+	pub fn raw_call(
```
AsceticBear, line 401: let this call ready-only truly.

HackFisher:
Is the internal_transact transactional, in other words, if the internal_transact failed, any of this following will happened:

1. Some event get deposited. (Expected Yes for now, looks ok?)
2. Sub state of the account in evm is modified. (Expected No for now)

AsceticBear:
> Some event get deposited. (Expected Yes for now, looks ok?)

Yes.

> Sub state of the account in evm is modified. (Expected No for now)

No. When the EVM executor exits with an error, the modified state will be reverted in the runner. Just like normal transactions.

frame/dvm/src/lib.rs
xiaoch05: how to resolve this?

frame/dvm/src/lib.rs
```diff
@@ -574,6 +511,117 @@ impl<T: Config> Pallet<T> {
...
+						)),
+						pays_fee: Pays::No,
+					})
+				}
+				ExitReason::Error(e) => {
+					log::error!("Executing internal transaction error happened, {:?}", e);
+					return Err(Error::<T>::InternalTransactionExitError.into());
+				}
+				ExitReason::Revert(e) => {
+					log::error!("Executing internal transaction error happened, {:?}", e);
```
xiaoch05, line 611: can this Revet exception be recorded in the ethereum transaction?

AsceticBear: Yes, they can be recorded in the event and receipt.

frame/dvm/src/tests.rs
```diff
@@ -520,6 +520,269 @@ fn call_should_handle_errors() {
...
+		// Check nonce
+		assert_eq!(
+			<Test as darwinia_evm::Config>::RingAccountBasic::account_basic(&INTERNAL_CALLER).nonce,
+			U256::from(2)
+		);
+	});
+}
+
+#[test]
+fn raw_call_with_insufficient_gas_limit_failed() {
```
AsceticBear, line 587: @xiaoch05  The test provides a clear explanation of `InternalTransactionGasLimit`.

bin/node/runtime/pangolin/src/pallets/dvm.rs
```diff
@@ -2,9 +2,14 @@
 use crate::*;
 use dvm_ethereum::{Config, IntermediateStateRoot};
 
+frame_support::parameter_types! {
+	pub InternalTransactionGasLimit: U256 = U256::from(300_000_000);
```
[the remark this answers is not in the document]
AsceticBear: This value can not be too low, which may cause internal transactions to fail due to out of gas.

HackFisher:
Let me rethink and review this week.

The first question is

Is internal trasaction must success? This looks quite different to dvm transaction. Just because it is called from internal, or how to catch the result if it failed?

frame/dvm/src/lib.rs
```diff
@@ -298,107 +300,25 @@ pub mod pallet {
...
 	pub fn internal_transact(target: H160, input: Vec<u8>) -> DispatchResultWithPostInfo {
 		ensure!(
 			dp_consensus::find_pre_log(&<frame_system::Pallet<T>>::digest()).is_err(),
 			Error::<T>::PreLogExists,
 		);
 		let nonce =
 			<T as darwinia_evm::Config>::RingAccountBasic::account_basic(&INTERNAL_CALLER).nonce;
-		let transaction = DVMTransaction::new(nonce, target, input);
-
+		let transaction = DVMTransaction::new_internal_transaction(
```
HackFisher, line 311: This dvm transaction looks wired

frame/dvm/src/lib.rs
```diff
@@ -574,6 +511,117 @@ impl<T: Config> Pallet<T> {
...
+			for topic in log.topics {
+				bloom.accrue(BloomInput::Raw(&topic[..]));
+			}
+		}
+	}
+
+	/// The dispatch result depends on the transaction type.
+	/// The internal transaction will return DispatchError when exitReason does not Succeed,
+	/// The Rpc transaction will alway return Ok(...) no matter what exitReason is.
+	pub fn return_dispatch_result(
```
HackFisher, line 591: Could we just directly return.reason in this method, and match and give different result according to the call context.

AsceticBear: No, we can't. This method return `DispatchResultWithPostInfo` and no concrete infomation.

```rs
pub type DispatchResultWithPostInfo = sp_runtime::DispatchResultWithInfo<crate::weights::PostDispatchInfo>;

pub struct PostDispatchInfo {
	/// Actual weight consumed by a call or `None` which stands for the worst case static weight.
	pub actual_weight: Option<Weight>,
	/// Whether this transaction should pay fees when all is said and done.
	pub pays_fee: Pays,
}
```

AsceticBear:
> Internal trasaction must success? This looks quite different to dvm transaction. Just because it is called from internal, or how to catch the result if it failed?

The things here are somehow complicated. Let me do more explanation.

1. Any transaction can be failed, both RPC transactions and internal transactions.

    - https://github.com/darwinia-network/darwinia-common/blob/fix-704-709/frame/dvm/src/lib.rs#L345
    - https://github.com/darwinia-network/darwinia-common/blob/fix-704-709/frame/evm/src/runner/stack.rs#L65

    When the transaction goes into the evm pallet runner from `dvm execute(...)` and the check fails before 
    entering evm executor. In this case, we consider the transaction executed failed, regardless of the type.

2. Two different strategies to handle evm executor exit error.

That's the key point of this pull request. 

- https://github.com/darwinia-network/darwinia-common/blob/fix-704-709/frame/evm/src/runner/stack.rs#L99
- https://github.com/darwinia-network/darwinia-common/blob/fix-704-709/frame/evm/src/runner/stack.rs#L142

Here, we assume that all checks before entering evm-executor are passed. The L142 shows the `runner execute(...)` will always return `Ok()` and do not care about the evm-executor `exit_reason`. Take a example, if a transaction encounters a `revert` error in the runner. In the view of outside dvm, the `Self::execute(..)` return is `ok()`.

xiaoch05:
What's the difference between dvm transaction and internal transaction?
I think it is:
dvm transaction DO NOT change state except dvm pallet(evm state), but internal transaction may change both evm state and other pallet's state. So dvm transaction don't need to consider the result from evm, and it always returns OK, and the API result is successes means that it has sent a transaction successfully. 
For internal transaction, if evm state is reverted, then all the states must be reverted.

xiaoch05:
@AsceticBear how do you solve the do_call read-only feature?

bin/node/runtime/pangolin/src/pallets/ethereum_issuing.rs
```diff
@@ -8,10 +8,15 @@ frame_support::parameter_types! {
 	pub const EthereumIssuingPalletId: PalletId = PalletId(*b"da/ethis");
 }
 
+frame_support::parameter_types! {
+	pub RawCallGasLimit: U256 = U256::from(300_000_000);
```
xiaoch05, line 12: there are so many these same definitions. Can it be defined at a common place?

AsceticBear: Initially, I want to make this gas value configurable. It is not necessary to do so at the moment.

AsceticBear: Fixed.

AsceticBear:
> What's the difference between dvm transaction and internal transaction?
I think it is:
dvm transaction DO NOT change state except dvm pallet(evm state), but internal transaction may change both evm state and other pallet's state. So dvm transaction don't need to consider the result from evm, and it always returns OK, and the API result is successes means that it has sent a transaction successfully.
For internal transaction, if evm state is reverted, then all the states must be reverted.

Yes, you are right.

> @AsceticBear how do you solve the do_call read-only feature?

Unfortunately, there is no way to restrict a pure read-only call inside the dvm pallet.

frame/dvm/Cargo.toml
```diff
@@ -15,8 +15,8 @@ codec          = { package = "parity-scale-codec", version = "2.1.1", default-fe
 ethereum       = { version = "0.7.1", default-features = false, features = ["with-codec"] }
 ethereum-types = { version = "0.11.0", default-features = false }
 evm            = { version = "0.27.0", default-features = false, features = ["with-codec"] }
-libsecp256k1 = { version = "0.5", default-features = false, features = ["static-context", "hmac"] }
-log            = { version = "0.4.14", optional = true }
+libsecp256k1   = { version = "0.5", default-features = false, features = ["static-context", "hmac"] }
+log            = { version = "0.4.14", default-features = false }
```
AurevoirXavier, line 19: `log` is no std. No need to `default-features = false`

AsceticBear:
```rs
failures:

---- tests::no_honesty_should_work stdout ----
thread 'tests::no_honesty_should_work' panicked at 'assertion failed: `(left == right)`
  left: `105`,
 right: `95`', frame/bridge/relayer-game/src/tests.rs:765:9
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

@AurevoirXavier Can you help me with this? This unit test failed as well in master ci.

frame/dvm/src/lib.rs
```diff
@@ -170,7 +168,9 @@ pub mod pallet {
 		InvalidSignature,
 		/// Pre-log is present, therefore transact is not allowed.
 		PreLogExists,
-		/// Call failed
+		/// The internal transaction failed
+		InternalEvmExecute,
```
AurevoirXavier, line 172: FailedToExecuteXXX would be better

frame/dvm/src/lib.rs
```diff
@@ -188,33 +189,31 @@ pub mod pallet {
...
 				}
-
-				let origin = Self::recover_signer(&transaction).ok_or_else(|| {
+				// Check signature correctly
+				let origin = recover_signer(&transaction).ok_or_else(|| {
 					InvalidTransaction::Custom(TransactionValidationError::InvalidSignature as u8)
 				})?;
-
+				// Check transaction gas limit correctly
 				if transaction.gas_limit >= T::BlockGasLimit::get() {
```
AurevoirXavier, line 197: Recommend to use `ensure!`

AsceticBear: Just keep the same with upstream.

frame/dvm/src/lib.rs
AsceticBear: Fixed

AurevoirXavier:
> ```rust
> failures:
> 
> ---- tests::no_honesty_should_work stdout ----
> thread 'tests::no_honesty_should_work' panicked at 'assertion failed: `(left == right)`
>   left: `105`,
>  right: `95`', frame/bridge/relayer-game/src/tests.rs:765:9
> note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
> ```
> 
> @AurevoirXavier Can you help me with this? This unit test failed as well in master ci.

It cannot reproduce locally.

I have seen this many times. Re-run would make it passed.

frame/bridge/ethereum/issuing/src/lib.rs
```diff
@@ -355,8 +355,7 @@ impl<T: Config> Pallet<T> {
 		let factory_address = MappingFactoryAddress::<T>::get();
 		let bytes = mtf::encode_mapping_token(backing, source)
 			.map_err(|_| Error::<T>::InvalidIssuingAccount)?;
-		let mapped_address = dvm_ethereum::Pallet::<T>::do_call(factory_address, bytes)
-			.map_err(|e| -> &'static str { e.into() })?;
+		let mapped_address = dvm_ethereum::Pallet::<T>::raw_call(factory_address, bytes)?;
```
HackFisher, line 358: `raw_call` is readonly, will it write some thing here?

frame/dvm/src/lib.rs
```diff
@@ -501,15 +413,36 @@ impl<T: Config> Pallet<T> {
...
+		let (_, _, info) = Self::execute(
+			INTERNAL_CALLER,
+			input,
+			U256::zero(),
+			U256::from(INTERNAL_TX_GAS_LIMIT),
+			None,
+			None,
+			TransactionAction::Call(contract),
+			None,
+		)?;
```
HackFisher, line 434: Could we make it readonly?

frame/bridge/ethereum/issuing/src/lib.rs
AsceticBear: No, just a read-only call to the smart contract and catch the return value.

frame/dvm/src/lib.rs
AsceticBear: This is not possible at the moment, the call here will directly call the executor in evm. Now, I added some docs here to avoid `raw_call` abuse.

HackFisher: Try this way, and add a unit test.
```
		let info = with_transaction(|| {
			let (_, _, info) = Self::execute(
				INTERNAL_CALLER,
				input.clone(),
				U256::zero(),
				U256::from(0x300000),
				None,
				None,
				TransactionAction::Call(contract),
				None,
			)?;

			Rollback(info)
		});
```
cc @AurevoirXavier @xiaoch05

AsceticBear: Fixed

AurevoirXavier:
The easiest way to check the storage changes is using `storage_root()`

AsceticBear:
> The easiest way to check the storage changes is using `storage_root()`

Certainly! Comparing storage root is simpler. The old assert is retained because it makes the test case much easier to understand.

HackFisher (approved):

frame/dvm/src/lib.rs
```diff
@@ -298,139 +297,52 @@ pub mod pallet {
...
+		Self::raw_transact(transaction).map(|(reason, used_gas)| match reason {
+			// Only when exit_reason is successful, return Ok(...)
+			ExitReason::Succeed(_) => Ok(PostDispatchInfo {
+				actual_weight: Some(T::GasWeightMapping::gas_to_weight(
+					used_gas.unique_saturated_into(),
+				)),
+				pays_fee: Pays::No,
+			}),
+			_ => {
+				log::error!("Executing internal transaction error happened");
```
AurevoirXavier, line 321: Still need this?

AsceticBear: Did you mean this log line or this match branch `_`?

AurevoirXavier: The log.

The error can be caught outside now.

AsceticBear: Just keep it.

PR #714, opened 2021-07-19
add mapping token factory setting

xiaoch05:


AurevoirXavier (changes requested):
Since you've introduced some new storage, please add the migration code that corresponds to them.

AsceticBear:
> Since you've introduced some new storage, please add the migration code that corresponds to them.

The s2s issuing pallet has not been deployed in the pangolin network? So no need to migrate.

xiaoch05:
yes this is the first time to deploy it.

AurevoirXavier:
> > Since you've introduced some new storage, please add the migration code that corresponds to them.
> 
> The s2s issuing pallet has not been deployed in the pangolin network? So no need to migrate.

So the `GenesisBuilder` had never set up the genesis storage; `build_genesis()` had never been executed.

AurevoirXavier:
`OnRuntimeUpgrade` would not trigger the `build_genesis()`.

AsceticBear:
O, you are right, we need a migration here.

AsceticBear:
I guess this is ready for review?

AsceticBear (approved):

AurevoirXavier:
Use constant instead of storage which means the address can only be modified on runtime upgrade. Is this expected?

xiaoch05:
> Use constant instead of storage which means the address can only be modified on runtime upgrade. Is this expected?

Yes, usually we're forbidden to modify this address. Instead, upgrade the logic contract to upgrade the version. Unless we have a fatal error we may modify this address by runtime upgrade.

HackFisher:
> Use constant instead of storage which means the address can only be modified on runtime upgrade. Is this expected?

I would suggest to use storage for the reason mentioned.

HackFisher:
Only sudo can change this strorage, which we can also say "forbidden".

xiaoch05:
> Only sudo can change this strorage, which we can also say "forbidden".

I agree with this suggestion. It's convenient to debug and deploy. We don't need to deploy the contract first for each network first, then upgrade runtime to match the address.

AsceticBear (approved):

HackFisher (approved):
No migration, but LGTM if using sudo set can also meet the requirement.

PR #718, opened 2021-07-22
Gas to weight = ?

### LINKED ISSUE #705: The ratio of the gas to weight
current we use 1:1 from gas to weight.
Suggest: 1:1000

AsceticBear:
Close #705 

This pull request answers a question this time in DVM.  1 gas =? weight

> From substrate.dev:
> Substrate defines one unit of weight as one picosecond of execution time, that is 10^12 weight = 1 second, or 1,000 weight = 1 nanosecond, on fixed reference hardware (Intel Core i7-7700K CPU with 64GB of RAM and an NVMe SSD).

According to the benchmark result, we set the radio of gas and weight 16_000.


frame/evm/src/benchmarking.rs
```diff
@@ -0,0 +1,119 @@
...
+		//     }
+
+		//     function infinite() public {
+		//         while (true) {
+		//             count=count+1;
+		//         }
+		//     }
+		// }
+
+		let contract_bytecode = hex::decode(concat!(
```
AurevoirXavier, line 58: `array_bytes::hex2bytes`

frame/evm/Cargo.toml
```diff
@@ -61,3 +65,8 @@ std = [
 	"sp-runtime/std",
 	"sp-std/std",
 ]
+
+runtime-benchmarks = [
+	"hex",
```
AurevoirXavier, line 70: drop

frame/evm/Cargo.toml
```diff
@@ -15,6 +15,7 @@ codec           = { package = "parity-scale-codec", version = "2.1.1", default-f
 evm             = { version = "0.27.0", default-features = false, features = ["with-codec"] }
 evm-gasometer   = { version = "0.27.0", default-features = false }
 evm-runtime     = { version = "0.27.0", default-features = false }
+hex             = { optional = true, version = "0.4", default-features = false }
```
AurevoirXavier, line 18: drop

frame/evm/src/benchmarking.rs
AurevoirXavier: And I think the `concat!` is unecessary?

HackFisher (approved):
LGTM

AurevoirXavier (approved):

PR #728, opened 2021-07-28
feat: add verify header fn

songtianyi:


frame/bridge/ethereum-bsc/src/lib.rs
```diff
@@ -354,6 +402,19 @@ pub mod pallet {
...
+		/// Verify single header
+		/// The header number should in the range [genesis_header.number, finalized_checkpoint.number + N]
+		/// Before the first call of verify_and_update_authority_set extrinsic, genesis_header == finalized_checkpoint
+		pub fn verify_header(header: &BSCHeader) -> DispatchResult {
+			let cfg = T::BSCConfiguration::get();
+			Self::contextless_checks(&cfg, header)?;
+			let authorities = <FinalizedAuthority<T>>::get();
+			// check signer
+			let signer = Self::recover_creator(cfg.chain_id, header)?;
+			ensure!(contains(&authorities, signer), <Error::<T>>::InvalidSigner);
```
xiaoch05, line 414: if the header is old. We can't verify it using the latest authority set.

songtianyi: Yes. So i add a new storage to track the older authorities. At least, we can ensure your header is signed by a valid authority.
BUT there are still some corner cases. 

Authority B was qualified in round 10, but not qualified at round 11, if you call this fn with a header signed by B and it's number is in round 11, this fn will return OK, but actually, the header is illegal

xiaoch05: we can't restrict the header number in a range. any header should be verified when user submits. 
I think all the authorities at every epoch should be stored.

songtianyi: Agree. But please note, the header's number must be greater than the genesis header, that's where we start to store the authority set.

xiaoch05: Yes the header older than genesis is not needed.

songtianyi: please review it again.

frame/bridge/ethereum-bsc/src/lib.rs
```diff
@@ -354,6 +431,36 @@ pub mod pallet {
...
+
+			// get index vec
+			let authorities_of_round = <AuthoritiesOfRound<T>>::get(round_key);
+			// get all authorities
+			let authorities = <Authorities<T>>::get();
+			// filter authorities of this round out
+			let signers = authorities_of_round
+				.iter()
+				.map(|i| authorities[*i as usize])
+				.collect::<Vec<Address>>();
```
xiaoch05, line 457: for each round, do we have the same authorities set? If not, we can't save the index instead of authorities.

songtianyi: No, the authorities' slice may be(or not) growing in each round. So the authorities are not always the same.
But all qualified authorities will be appended to the set, so it's ok to have an index, not the address itself.
The only thing we should care about is to make sure the order is stable.

xiaoch05: ok I understand, authorities set is an array that saves all the authorities currently active or appeared in history. and the index to the authority is stable so we can find the authority by its index. Then for each round, we only need to record indexes of the authorities.

frame/bridge/ethereum-bsc/src/lib.rs
```diff
@@ -354,6 +431,36 @@ pub mod pallet {
 			Ok(signers)
 		}
 
+		/// Verify single header
+		/// The header number should in the range [genesis_header.number, finalized_checkpoint.number + N]
+		/// Before the first call of verify_and_update_authority_set extrinsic, genesis_header == finalized_checkpoint
```
AurevoirXavier, line 436: What does this line mean

AurevoirXavier:
Does this come from the go code? If yes, could you link the code here?

AurevoirXavier (approved):

frame/bridge/ethereum-bsc/src/lib.rs
songtianyi: 意思是 最初始的时候，genesis header 就是 finalized header

songtianyi:
> Does this come from the go code? If yes, could you link the code here?

No, it's a general method.

HackFisher (approved):

PR #730, opened 2021-07-29
Code Cleaning

AurevoirXavier:
Restruct the cli part args for the future service refactor.

<sup> Created from VS Code using [CodeStream](https://codestream.com/?utm_source=cs&utm_medium=pr&utm_campaign=github*com)</sup>


bin/node/cli/src/cli.rs
```diff
@@ -62,13 +62,9 @@ pub struct RunCmd {
...
 	pub authority_discovery_disabled: bool,
 
-	/// Maximum number of logs in a query.
-	#[structopt(long, default_value = "10000")]
-	pub max_past_logs: u32,
-
-	/// The dynamic-fee pallet target gas price set by block author
-	#[structopt(long, default_value = "1000000000")]
-	pub target_gas_price: u64,
+	#[allow(missing_docs)]
```
AsceticBear, line 65: Remove this allow and add some documentation instead.

bin/node/cli/src/cli.rs
```diff
@@ -117,3 +113,14 @@ pub enum Subcommand {
...
+#[derive(Debug, StructOpt)]
+pub struct DynamicFeeParameters {
+	/// Maximum number of logs in a query.
+	#[structopt(long, default_value = "10000")]
+	pub max_past_logs: u32,
+
+	/// The dynamic-fee pallet target gas price set by block author
+	#[structopt(long, default_value = "1000000000")]
+	pub target_gas_price: u64,
+}
```
AsceticBear, line 126: Actually, `max_past_logs` has nothing to do with DynamicFee. I think to keep the old way would be better.

bin/node/cli/src/cli.rs
AurevoirXavier: This is a flattened field. The user can not see this from the cli.

bin/node/cli/src/cli.rs
AurevoirXavier: Rename to `DvmArgs`.

AsceticBear: Still not push yet?

bin/node/cli/src/cli.rs
```diff
@@ -62,13 +62,9 @@ pub struct RunCmd {
...
-	/// Maximum number of logs in a query.
-	#[structopt(long, default_value = "10000")]
-	pub max_past_logs: u32,
-
-	/// The dynamic-fee pallet target gas price set by block author
-	#[structopt(long, default_value = "1000000000")]
-	pub target_gas_price: u64,
+	#[allow(missing_docs)]
+	#[structopt(flatten)]
+	pub dynamic_fee_parameters: DvmArgs,
```
AsceticBear, line 67: pub dvm_args: DvmArgs,

AsceticBear (changes requested):

HackFisher (approved):

PR #731, opened 2021-08-03
refactor: ethereum to darwinia erc20 bridge

xiaoch05:
This PR includes these items

- [x] Move contract encoder/decoder from issuing pallet to primitives. Each contract has its own decoder mod.
- [x] Remove IssuingHandle from EVM pallet. Instead, using dispatch precompile to receive messages from the contract.
- [x] Add two setting interfaces for mapping-token-factory address and backing address.
- [x] Benchmark test and weight info for ethereum issuing calls.


frame/bridge/ethereum/issuing/src/lib.rs
```diff
@@ -21,275 +21,351 @@
...
-		RegisterErc20(AccountId, EthereumAddress, EthereumTransactionIndex),
+		RegisterErc20(EthereumAddress, EthereumTransactionIndex),
 		/// redeem erc20 token
-		RedeemErc20(AccountId, EthereumAddress, EthereumTransactionIndex),
-		/// erc20 created
-		CreateErc20(EthereumAddress),
+		RedeemErc20(EthereumAddress, EthereumTransactionIndex),
 		/// burn event
 		/// type: 1, backing, sender, recipient, source, target, value
-		BurnToken(u8, EthereumAddress, EthereumAddress, EthereumAddress, EthereumAddress, EthereumAddress, U256),
```
AsceticBear, line 123: It's would be better to  add more docs about this type meaning. 
```rs
pub enum Token {
	InvalidToken,
	Native(TokenInfo),
	Erc20(TokenInfo),
}
```

frame/bridge/ethereum/issuing/src/lib.rs
```diff
@@ -21,275 +21,351 @@
...
-	where
-		AccountId = AccountId<T>,
-	{
+	#[pallet::event]
+	#[pallet::generate_deposit(pub(super) fn deposit_event)]
+	#[pallet::metadata(T::AccountId = "AccountId")]
+	pub enum Event<T: Config> {
 		/// register new erc20 token
-		RegisterErc20(AccountId, EthereumAddress, EthereumTransactionIndex),
+		RegisterErc20(EthereumAddress, EthereumTransactionIndex),
```
AsceticBear, line 119: Can we use `TokenRegistered(u8, EthereumAddress, EthereumAddress, EthereumAddress)` to replace `RegisterErc20`, I see there is a type field in the `TokenRegistered`  event.

frame/bridge/ethereum/issuing/src/lib.rs
```diff
@@ -21,275 +21,351 @@
...
+			EthereumAddress,
+			EthereumAddress,
+			EthereumAddress,
+			EthereumAddress,
+			EthereumAddress,
+			U256,
+		),
 		/// token registered event
 		/// type: u8 = 0, backing, source(origin erc20), target(mapped erc20)
 		TokenRegistered(u8, EthereumAddress, EthereumAddress, EthereumAddress),
```
AsceticBear, line 135: More docs about the `type` meaning?

frame/bridge/ethereum/issuing/src/lib.rs
xiaoch05: No，`RegisterErc20` and `TokenRegistered` are two different events. `TokenRegistered`  is deposited when the token registration is finished.

frame/bridge/ethereum/issuing/src/lib.rs
```diff
@@ -21,275 +21,351 @@
...
+			Ok(().into())
 		}
 	}
 }
 
-impl<T: Config> Module<T> {
+impl<T: Config> Pallet<T> {
 	/// The account ID of the issuing pot.
-	pub fn dvm_account_id() -> H160 {
+	pub fn dvm_account_id() -> EthereumAddress {
```
AsceticBear, line 333: Is this function no longer in use? Compile won't fire when I remove it.

frame/bridge/ethereum/issuing/src/lib.rs
```diff
@@ -21,275 +21,351 @@
...
 
-decl_error! {
+	#[pallet::error]
 	/// Issuing pallet errors.
-	pub enum Error for Module<T: Config> {
+	pub enum Error<T> {
 		/// Invalid Issuing System Account
 		InvalidIssuingAccount,
 		/// assert ar
 		AssetAR,
```
AsceticBear, line 99: `AR` means `Assert Already Registered`? Maybe we should give a more understandable name.

frame/bridge/ethereum/issuing/src/lib.rs
AsceticBear: `RegisterErc20` says the first step of registering token, `TokenRegistered` says the registration process successfully?

xiaoch05: Yes

frame/bridge/ethereum/issuing/src/lib.rs
xiaoch05: resolved. https://github.com/darwinia-network/darwinia-common/compare/b23ee2338082...864e1726f093

frame/bridge/ethereum/issuing/src/lib.rs
```diff
@@ -21,275 +21,347 @@
...
+		#[pallet::weight(<T as Config>::WeightInfo::mapping_factory_event_handle())]
+		pub fn mapping_factory_event_handle(
+			origin: OriginFor<T>,
+			input: Vec<u8>,
+		) -> DispatchResultWithPostInfo {
+			let caller = ensure_signed(origin)?;
+			ensure!(input.len() >= 8, <Error<T>>::InvalidInput);
+			let factory = MappingFactoryAddress::<T>::get();
+			let factory_id = <T as darwinia_evm::Config>::AddressMapping::into_account_id(factory);
+			ensure!(factory_id == caller, <Error<T>>::AssetAR);
```
AsceticBear, line 284: `<Error<T>>::AssetAR)` This error is inaccurate here?  May something like `<Error<T>>::BadCaller)`

frame/bridge/ethereum/issuing/src/lib.rs
```diff
@@ -21,275 +21,347 @@
...
+					burn_info.source,
+					EthereumAddress::from_slice(burn_info.recipient.as_slice()),
+					burn_info.amount,
+				)?;
+			} else if &input[4..8] == register_action {
+				let register_info = TokenRegisterInfo::decode(&input[8..])
+					.map_err(|_| Error::<T>::InvalidInputData)?;
+				Self::finish_token_registered(register_info.0, register_info.1, register_info.2);
+			} else {
+				log::trace!("Unsupport action!");
```
AsceticBear, line 303: Return an error instead of log a message?

frame/bridge/ethereum/issuing/src/lib.rs
```diff
@@ -21,275 +21,347 @@
...
+		))
 	}
 
 	pub fn finish_token_registered(
 		backing: EthereumAddress,
 		source: EthereumAddress,
 		target: EthereumAddress,
 	) -> DispatchResult {
-		let raw_event = RawEvent::TokenRegistered(0, backing, source, target);
+		let raw_event = Event::TokenRegistered(0, backing, source, target);
```
AsceticBear, line 371: `Event::TokenRegistered(0, backing, source, target);` Make this `0` to a const!!

frame/bridge/ethereum/issuing/src/lib.rs
xiaoch05: We can't return an error here. If so, the contract will be reverted.

frame/bridge/ethereum/issuing/src/lib.rs
AsceticBear: Good name.

frame/bridge/ethereum/issuing/src/mock_header.rs
```diff
@@ -0,0 +1,13 @@
...
...000000000008000000000000000000000000000000000000000000000042000000000000000000000000000000000000000000000010000000000000000000000000000000000000000000000000200000000000080000000000000000000000000000000008000000000000000000800000000800000000000000000002000000000000000000000000004000000000000000000000000000000000200000100000000000000000000000000000000000000000000000000000f9019af89b94f2d4cafe299a997127915d2e44afbc86ed7082a2f863a0ddf252ad1be2c89b69c2b068fc378daa952ba7f163c4a11628f55a4df523b3efa0000000000000000000000000f2d4cafe299a997127915d2e44afbc86ed7082a2a0000000000000000000000000b2bea2358d817dae01b0fd0dc3aecb25910e65aaa00000000000000000000000000000000000000000000000000000000000000064f8fb94b2bea2358d817dae01b0fd0dc3aecb25910e65aaf842a0c2088d2d8c703cc9316b03c6aeba47f60b71c0a979c6c8ee5d60c9143511dfa5a0000000000000000000000000f2d4cafe299a997127915d2e44afbc86ed7082a2b8a0000000000000000000000000f2d4cafe299a997127915d2e44afbc86ed7082a2000000000000000000000000c1332ecce3a065bfe8babd8d28ee0b1a7c76cbcc0000000000000000000000000000000000000000000000000000000000000064000000000000000000000000d2c7008400f54aa70af01cf8c747a4473246593e0000000000000000000000000000000000000000000000000000000000000000";
```
AsceticBear, line 13: How about putting these mock data to the `benchmarking.rs`, since the things are only used in benchmark now.

xiaoch05: It is better to put these mock data in a single file. 
The data is too large.
It will cause large memory usage and response slowly for the editor.

HackFisher (approved):

AurevoirXavier:
Companion for darwinia: https://github.com/darwinia-network/darwinia/pull/713

PR #742, opened 2021-08-12
Add `transactional` attribute to some call

### LINKED ISSUE #739: Ensure call which interact with internal transaction should be transactional.

AsceticBear:
Close #739

### DIFF
--- a/frame/bridge/s2s/issuing/src/lib.rs
+++ b/frame/bridge/s2s/issuing/src/lib.rs
@@ -36,7 +36,7 @@ use frame_support::{
 	ensure,
 	pallet_prelude::*,
 	traits::{Currency, ExistenceRequirement::*, Get},
-	PalletId,
+	transactional, PalletId,
 };
 use frame_system::ensure_signed;
 use sp_runtime::{traits::Convert, DispatchError, SaturatedConversion};
@@ -102,7 +102,7 @@ pub mod pallet {
 		#[pallet::weight(
 			<T as Config>::WeightInfo::asset_burn_event_handle()
 		)]
-		#[frame_support::transactional]
+		#[transactional]
 		pub fn asset_burn_event_handle(
 			origin: OriginFor<T>,
 			input: Vec<u8>,
@@ -147,6 +147,7 @@ pub mod pallet {
 			<T as Config>::WeightInfo::register_from_remote()
 			.saturating_add(2_000_000 * 3)
 		)]
+		#[transactional]
 		pub fn register_from_remote(
 			origin: OriginFor<T>,
 			token: Token,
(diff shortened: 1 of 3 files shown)


xiaoch05:
@AsceticBear why add these `transactional` attribute?
we don't need to do that, the call will not change any storage state expect evm state.

AsceticBear:
> the call will not change any storage state expect evm state.

Yes, that's true. However, we need to ensure that the changed storage in other pallets reverts when some error happened in the internal transaction(evm exectuion).

frame/bridge/s2s/issuing/src/lib.rs
```diff
@@ -102,7 +102,7 @@ pub mod pallet {
 		#[pallet::weight(
 			<T as Config>::WeightInfo::asset_burn_event_handle()
 		)]
-		#[frame_support::transactional]
+		#[transactional]
```
AsceticBear, line 105: This one is necessary, right? Since we perform a `transfer` before `burn_and_remote_unlock`.

xiaoch05: yes

xiaoch05:
> > the call will not change any storage state expect evm state.
> 
> Yes, that's true. ~However, we need to ensure that the changed storage in other pallets reverts when some error happened in the internal transaction(evm exectuion).~
> 
> I check again these calls, as you say, some are unnecessary. In my view, it would be much safer to add this attribute because these calls always have relations with other pallets, the logic is somehow complicated. Adding `[transactional]` here has no extra bad influence, but can prevent some potential issues in the future.

I can't agree with this. The call's logic is clear. If we add this unnecessary attribute, users will always recheck why transactional attribute here. And do this attribute cause more weight cost?

AsceticBear:
@HackFisher @AurevoirXavier  Any suggestions?

HackFisher:
which kind of dispatch call should be transactional(success or fail without state change)?

This PR LGTM, the dvm internal call will result error if fail right, may be should merge after #713？

AsceticBear:
> which kind of dispatch call should be transactional(success or fail without state change)?

https://substrate.dev/docs/en/knowledgebase/runtime/storage#verify-first-write-last

A example call like this:

```rs
// Pallet A
pub fn call_a1(origin: OriginFor<T>, ...) -> DispatchResultWithPostInfo {
	let user = ensure_signed(origin)?;
        Self::call_a2()?;   // this call change A's runtime storage

        B::call_b()?   // If this call failed and returns, we should ensure that the storage modified in `call_a2` is reverted.
        
	Ok(().into())
}
```

> the dvm internal call will result error if fail, right？ may be should merge after #713？

Internal calls or transacts return an error after 713.

HackFisher (approved):

PR #745, opened 2021-08-13
Drop `darwinia-cli`

AurevoirXavier:
Hard to maintain, and I found there is nobody is using this feature. 

If needed, we can open a PR for substrate later.


AsceticBear:
I am curious about the reason why we add `darwinia-cli` before?

AurevoirXavier:
> I am curious about the reason why we add `darwinia-cli` before?

https://github.com/darwinia-network/darwinia/issues/427

bin/node/cli/src/command.rs
```diff
@@ -111,7 +97,9 @@ pub fn run() -> sc_cli::Result<()> {
 
 	match &cli.subcommand {
 		None => {
-			let runner = Configuration::create_runner(Cli::from_args())?;
+			let runner = cli
+				.create_runner(&cli.run.base)
+				.map_err(sc_cli::Error::from)?;
```
AsceticBear, line 102: How about `create_runner(&cli.run.base)?;`?

PR #746, opened 2021-08-13
BEEFY integration

hujw77:


AurevoirXavier:
Need help?

HackFisher:
@hujw77 @xiaoch05 Prefer to use https://github.com/paritytech/grandpa-bridge-gadget (Or its fork if necessary) as dependency instead of copy to the common here.

hujw77:
> Need help?

https://github.com/paritytech/substrate/blob/287b5c5a8fe47beacfdfdf764b632b897691e9da/primitives/core/src/ecdsa.rs#L576
BEEFY use this function to verify sign, It needs the latest substrate version, Darwinia's substrate lib may need to upgrade.

hujw77:
> @hujw77 @xiaoch05 Prefer to use https://github.com/paritytech/grandpa-bridge-gadget (Or its fork if necessary) as dependency instead of copy to the common here.

OK

AurevoirXavier:
We fall behind the latest substrate code around 5-6 times runtime upgrade

AurevoirXavier:
Move to https://github.com/darwinia-network/darwinia-common/tree/xavier-beefy

PR #753, opened 2021-08-17
Managing ERC20 tokens from Runtime/Treasury

AsceticBear:


frame/bridge/subswap/Cargo.toml
```diff
@@ -0,0 +1,35 @@
+[package]
+authors     = ["Darwinia Network <hello@darwinia.network>"]
+description = "Interact with Sub Swap"
+edition     = "2018"
+homepage    = "https://darwinia.network/"
+license     = "GPL-3.0"
+name        = "darwinia-subswap"
```
HackFisher, line 7: This pallet should not be called as `subswap`, subswap is a thirdparty brand not owned by Darwinia.

To simplify this, I think we can use a smart contract as a helper to execute the operations. This pallet only need to support transfer some balances to a DVM Account, which as my understanding, not even need a pallet.

cc @xiaoch05 @hujw77

HackFisher:
One key question will be how treasury can hold ERC20 tokens since treasury's account is substrate account, not evm account.

How to generating proposal of `Transfering ERC20 Tokens`?

First, there must be some precompile contract holding these ERC20 Tokens, and can only managed by runtime, `Transfer` Precompile might be a good candidate.

Then, we need to support a dispatch call which can be used to describe this proposal, this dispatch call might require special origin(Root, or Council), and able to do an internal transaction to transfer the ERC20 Token in the precompile contract.

AsceticBear:
Another way is that we can use ERC20 deployed contract account to hold tokens and store the addresses in a pallet.
In addition, the pallet has dispatch calls that can be controlled by Root, or the Technical Committee, used to operate these tokens.

HackFisher:
We can be more simpler: using 0x0 as the address for runtime(and treasury), and add a dispatch call in dvm pallet for sending internal call to ERC20 contract to transfer the tokens it owns. (internal call from runtime can overpass signature checking, and call target contract as 0x0 as my understanding)

If we want to manage ERC20 tokens for runtime/treasury, we only need to transfer the tokens to 0x0. Then the dispatch call will be able to transfer it(With permissions from ROOT/Council).

xiaoch05:
> We can be more simpler: using 0x0 as the address for runtime(and treasury), and add a dispatch call in dvm pallet for sending internal call to ERC20 contract to transfer the tokens it owns. (internal call from runtime can overpass signature checking, and call target contract using 0x0 as `msg.sender` as my understanding)
> 
> If we want to manage ERC20 tokens for runtime/treasury, we only need to transfer the tokens to 0x0. Then the dispatch call will be able to transfer it(With permissions from ROOT/Council).

Yes we can do this like mapping-token-factory, all these interface can only be called by a special account `0x0` 
https://github.com/darwinia-network/darwinia-bridge-sol/blob/master/contracts/mapping-token/contracts/darwinia/DarwiniaMappingTokenFactory.sol#L43

hujw77:
0x0 is generally used in the contract Mint or Burn, and it's strange to transfer the token directly from 0x0.

frame/bridge/subswap/Cargo.toml
xiaoch05: The contract does all the things, such as add and removes liquidity, send ERC20 tokens or native tokens. And the owner of this contract is System account `0x0`. I think the ERC20 is not saved in the pallet, instead, it's saved in this contract.

xiaoch05:
> 0x0 is generally used in the contract Mint or Burn, and it's strange to transfer the token directly from 0x0.

The token is not transferred from `0x0` to other addresses. Instead, only `0x0` address has the right to send tokens from the  special contract's address to other addresses.

hujw77:
I think 0x0 is a default special address. If 0x0 has such a big right, it can call all the authority contract that is not initialized.

HackFisher:
Substrate default account for Pallet Id
PalletId -> Substrate Account "modlda/trsry" // 32 bytes

DVM acounnt for Pallet Id
PalletId -> EVMAddress "modlda/trsry" // 20 bytes -> EVM Derived Substrate Account "dvm:" + 0x00000000000000 + "modal/trsy" // 32bytes

Using for each dvm
into_dvm_address() -> EthreumAddress

Spec Constraint: https://github.com/darwinia-network/darwinia/wiki/Darwinia-Address-Format-Overview

Darwinia DVM address derived Substratea account，must use"dvm:" + 0x00000000000000 as Prefix

PR #763, opened 2021-08-18
Use `ensure_source_root` in `darwinia-support`

AsceticBear:


frame/support/src/lib.rs
```diff
@@ -83,12 +86,18 @@ pub mod s2s {
...
+	pub fn ensure_source_root<AccountId, Converter>(
 		chain_id: ChainId,
-	) -> AccountId {
+		account: &AccountId,
+	) -> Result<(), DispatchError>
+	where
+		AccountId: Encode + sp_std::cmp::PartialEq,
+		Converter: Convert<H256, AccountId>,
+	{
 		let hex_id = derive_account_id::<AccountId>(chain_id, SourceAccount::Root);
```
AurevoirXavier, line 97: It seems `SouceAccount` is not in this scope.

frame/support/src/lib.rs
```diff
@@ -83,12 +86,18 @@ pub mod s2s {
...
 	}
 
-	pub fn source_root_converted_id<AccountId: Encode, Converter: Convert<H256, AccountId>>(
+	pub fn ensure_source_root<AccountId, Converter>(
 		chain_id: ChainId,
-	) -> AccountId {
+		account: &AccountId,
+	) -> Result<(), DispatchError>
+	where
+		AccountId: Encode + sp_std::cmp::PartialEq,
```
AurevoirXavier, line 94:
```suggestion
	AccountId: PartialEq + Encode,
```

AsceticBear: Fixed.

frame/support/src/lib.rs
```diff
@@ -31,7 +31,7 @@ pub mod balance {
 }
 
 pub mod evm {
-	// --- darwinia ---
+	// --- crates.io ---
 	use ethereum::TransactionMessage;
 	use ethereum_primitives::{H160, H256};
```
AurevoirXavier, line 36: darwinia part

frame/support/src/lib.rs
```diff
@@ -59,12 +59,18 @@ pub mod evm {
 
 // TODO: Should we move this to `s2s-primitives`?
 pub mod s2s {
-	use bp_runtime::{derive_account_id, ChainId, SourceAccount};
+	// --- crates.io ---
 	use codec::Encode;
 	use ethabi::{encode, Token};
 	use ethereum_primitives::{H160, H256};
```
AurevoirXavier, line 65: darwinia part

PR #765, opened 2021-08-19
Add `IntoDvmAddress` trait support

AsceticBear:


bin/node/runtime/pangolin/src/pallets/ethereum_issuing.rs
```diff
@@ -9,7 +9,8 @@ frame_support::parameter_types! {
 }
 
 impl Config for Runtime {
-	type PalletId = EthereumIssuingPalletId;
+	// FIXME: Remove tight couple with dvm_ethereum and change it to PalletId
+	type IssuingPalletId = EthereumIssuingPalletId;
```
HackFisher, line 13: why not use `PalletId` since it is convention?

frame/support/src/evm.rs
```diff
@@ -0,0 +1,122 @@
...
+pub struct ContractId(pub [u8; 8]);
+
+/// Convert from contract id to dvm address
+impl IntoDvmAddress for ContractId {
+	fn into_dvm_address(&self) -> H160 {
+		let mut bytes = vec![0u8; 12];
+		bytes.append(&mut self.0.to_vec());
+		H160::from_slice(&bytes)
+	}
+}
```
HackFisher, line 51: Suggest to not adding new Types since PalletId can meet the requirements

HackFisher (changes requested):

bin/node/runtime/pangolin/src/pallets/ethereum_issuing.rs
AsceticBear: The `ethereum_issuing` has a tight couple with the `dvm_ethereum` pallet Config.  Once we gave `dvm_ethereum` pallet a config field `PalletId`, the `ethereum_issuing` pallet will show the following compile error:
```rs
pub trait Config: dvm_ethereum::Config {
 ....
}
```
```sh
error[E0221]: ambiguous associated type `PalletId` in bounds of `T`
  --> frame/bridge/ethereum/issuing/src/lib.rs:76:1
   |
76 | #[frame_support::pallet]
   | ^^^^^^^^^^^^^^^^^^^^^^^^
   | |
   | ambiguous associated type `PalletId`
   | help: use fully qualified syntax to disambiguate: `<T as pallet::Config>::PalletId`
...
86 |         type PalletId: Get<PalletId>;
   |         ----------------------------- ambiguous `PalletId` from `pallet::Config`
```

A simpler way for this pr to solve this problem is renaming `PalletId` in *_issuing pallet. I will raise another pull request to release the couple between *_issuing pallet and dvm_ethereum pallet.

frame/support/src/evm.rs
AsceticBear: In my mind, `PalletId` is designed to generate a special module account id for pallets to identify a unique pallet account.

What we need is to generate a contract id instead of a pallet module id. Thus, I think directly use `PalletId` in runtime config might be confusing to others. Try to imagine that a pallet with two config type classes is PalletId.

bin/node/runtime/pangolin/src/pallets/ethereum_issuing.rs
HackFisher:
```
	pub trait Config: dvm_ethereum::Config {
		type Event: From<Event<Self>> + IsType<<Self as frame_system::Config>::Event>;
		#[pallet::constant]
		type PalletId: Get<PalletId>;
		type EthereumRelay: EthereumReceipt<Self::AccountId, RingBalance<Self>>;
		type EcdsaAuthorities: RelayAuthorityProtocol<Self::BlockNumber, Signer = EthereumAddress>;
		type WeightInfo: WeightInfo;
	}
```

@xiaoch05 It is confusing that why ethereum_issuing::Config has to inherit `dvm_ethereum::Config`, prefer to fix this FIXME in the same PR rather delay to future PR.

frame/bridge/s2s/backing/src/lib.rs
```diff
@@ -46,9 +48,12 @@ use sp_runtime::{
 };
 use sp_std::prelude::*;
 // --- darwinia ---
-use darwinia_support::s2s::{
-	ensure_source_root, to_bytes32, RelayMessageCaller, BACK_ERC20_RING, RING_DECIMAL, RING_NAME,
-	RING_SYMBOL,
+use darwinia_support::{
+	balance::*,
```
HackFisher, line 52: Check warning

bin/node/runtime/pangolin/src/pallets/ethereum_issuing.rs
AsceticBear: OKay. I will push the fix later.

frame/support/src/evm.rs
HackFisher: As the name hints, let's define the PalletId as a unique ID with type `[u8; 8]` representing this pallet, substrate account and DVM account are generated from the same pallet id.

Adding extra types add more concept and complexity, prefer to keep it simple

AsceticBear:
@HackFisher @AurevoirXavier @xiaoch05  Review needed.

frame/support/src/evm.rs
```diff
@@ -0,0 +1,109 @@
...
+pub trait IntoDvmAddress {
+	fn into_dvm_address(&self) -> H160;
+}
+
+// Convert pallet id to dvm address
+impl IntoDvmAddress for PalletId {
+	fn into_dvm_address(&self) -> H160 {
+		let mut bytes = vec![0u8; 12];
+		bytes.append(&mut self.0.to_vec());
+		H160::from_slice(&bytes)
```
HackFisher, line 45: What's the H160 looks like?

It looks it will have prefix with 12 0u8, [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, self's elements ] , which is wired.

Document the spec here, and refer the into_sub_account of PalletId:

```
/// Format is TYPE_ID ++ encode(parachain ID) ++ 00.... where 00... is indefinite trailing zeroes to
/// fill AccountId.
impl<T: Encode + Decode + Default, Id: Encode + Decode + TypeId> AccountIdConversion<T> for Id {
	fn into_sub_account<S: Encode>(&self, sub: S) -> T {
		(Id::TYPE_ID, self, sub)
			.using_encoded(|b| T::decode(&mut TrailingZeroInput(b)))
			.unwrap_or_default()
	}

	fn try_from_sub_account<S: Decode>(x: &T) -> Option<(Self, S)> {
		x.using_encoded(|d| {
			if &d[0..4] != Id::TYPE_ID {
				return None
			}
			let mut cursor = &d[4..];
			let result = Decode::decode(&mut cursor).ok()?;
			if cursor.iter().all(|x| *x == 0) {
				Some(result)
			} else {
				None
			}
		})
	}
}
```

HackFisher (changes requested):

frame/support/src/evm.rs
HackFisher: https://doc.rust-lang.org/std/vec/struct.Vec.html#method.append

frame/bridge/ethereum/issuing/src/lib.rs
```diff
@@ -79,13 +79,15 @@ pub mod pallet {
 
 	#[pallet::config]
 	#[pallet::disable_frame_system_supertrait_check]
-	pub trait Config: dvm_ethereum::Config {
-		type Event: From<Event<Self>> + IsType<<Self as frame_system::Config>::Event>;
+	pub trait Config: frame_system::Config + darwinia_evm::Config {
```
xiaoch05, line 82: can we not use  darwinia_evm::Config?

bin/node/runtime/pangolin/src/pallets/ethereum_issuing.rs
```diff
@@ -11,7 +11,9 @@ frame_support::parameter_types! {
 impl Config for Runtime {
 	type PalletId = EthereumIssuingPalletId;
 	type Event = Event;
+	type RingCurrency = Ring;
 	type EthereumRelay = EthereumRelay;
 	type EcdsaAuthorities = EthereumRelayAuthorities;
 	type WeightInfo = ();
+	type InternalTransactHandler = Ethereum;
```
xiaoch05, line 18: rename InternalTransactHandler

frame/bridge/s2s/backing/src/lib.rs
```diff
@@ -69,9 +73,12 @@ pub mod pallet {
 	pub trait Config: frame_system::Config {
 		type Event: From<Event<Self>> + IsType<<Self as frame_system::Config>::Event>;
 		type WeightInfo: WeightInfo;
+
 		#[pallet::constant]
 		type PalletId: Get<PalletId>;
 		#[pallet::constant]
+		type BackErc20RingId: Get<PalletId>;
```
xiaoch05, line 80:
```suggestion
		type Erc20RingId: Get<PalletId>;
```

frame/support/src/evm.rs
```diff
@@ -0,0 +1,109 @@
...
+		target: H160,
+		input: Vec<u8>,
+	) -> Self {
+		let transaction = ethereum::Transaction {
+			nonce,
+			// Not used, and will be overwritten by None later.
+			gas_price: U256::zero(),
+			gas_limit: U256::from(INTERNAL_TX_GAS_LIMIT),
+			action: ethereum::TransactionAction::Call(target),
+			value: U256::zero(),
```
xiaoch05, line 90: we may need use this value in future.

bin/node/runtime/pangolin/src/pallets/ethereum_issuing.rs
AsceticBear: any suggestion?

frame/bridge/ethereum/issuing/src/lib.rs
AsceticBear: Yes, we can remove the `darwinia_evm::Config` bind here. However, I don't think this is necessary now, the darwinia_evm config is simple.

frame/bridge/s2s/backing/src/tests.rs
```diff
@@ -0,0 +1,179 @@
...
+		.unwrap();
+	t.into()
+}
+
+#[test]
+fn test_back_erc20_dvm_address() {
+	new_test_ext().execute_with(|| {
+		assert_eq!(
+			<Test as s2s_backing::Config>::BackErc20RingId::get().into_dvm_address(),
+			EthereumAddress::from_str("0x00000000000000000000000064612f7332736272").unwrap()
```
AsceticBear, line 166: @HackFisher The generated address looks like this.

frame/support/src/evm.rs
AsceticBear: I change the `into_dvm_address` impl. Please check again.

frame/support/src/evm.rs
AsceticBear: If this value is needed in the future, take it out at that time.

frame/bridge/s2s/backing/src/lib.rs
```diff
@@ -69,9 +73,12 @@ pub mod pallet {
 	pub trait Config: frame_system::Config {
 		type Event: From<Event<Self>> + IsType<<Self as frame_system::Config>::Event>;
 		type WeightInfo: WeightInfo;
+
 		#[pallet::constant]
 		type PalletId: Get<PalletId>;
 		#[pallet::constant]
+		type Erc20RingId: Get<PalletId>;
```
HackFisher, line 80:
```suggestion
		type RingPalletId: Get<PalletId>;
```

frame/bridge/s2s/backing/src/tests.rs
```diff
@@ -0,0 +1,179 @@
...
+	type OnKilledAccount = ();
+	type SystemWeightInfo = ();
+	type SS58Prefix = ();
+	type OnSetCode = ();
+}
+
+frame_support::parameter_types! {
+	pub const MockChainId: [u8; 4] = [0; 4];
+	pub const MockId: PalletId = PalletId(*b"da/s2sba");
+	pub const Erc20RingId: PalletId = PalletId(*b"da/s2sbr");
```
HackFisher, line 96:
```suggestion
	pub const RingPalletId: PalletId = PalletId(*b"da/bring");
```

HackFisher (changes requested):

frame/dvm/src/tests.rs
```diff
@@ -1370,3 +1374,14 @@ fn mutate_account_dec_balance_existential_by_990() {
...
 	});
 }
+
+#[test]
+fn test_pallet_id_to_dvm_address() {
+	let (_, mut ext) = new_test_ext(1);
+	ext.execute_with(|| {
+		assert_eq!(
+			<Test as self::Config>::PalletId::get().into_dvm_address(),
+			H160::from_str("0x0000000000000000000000006461722f64766d70").unwrap()
```
HackFisher, line 1384:
```suggestion
			H160::from_str("0x6d6f646c6461722f64766d700000000000000000").unwrap()
```

HackFisher (approved):

AurevoirXavier:
Companion: https://github.com/darwinia-network/darwinia/pull/723

PR #770, opened 2021-08-20
Support New Receipt Type from EIP-2718, EIP-1559

AurevoirXavier:


HackFisher (approved):

AurevoirXavier:
Interesting. It seems the types of changes get revert.

I'll update the types in another PR.

PR #782, opened 2021-08-30
New Bridge Name Spec

AurevoirXavier:
https://github.com/darwinia-network/darwinia/issues/714
https://github.com/darwinia-network/darwinia/issues/720


node/runtime/pangolin/Cargo.toml
```diff
@@ -58,6 +57,7 @@ pangolin-constants                  = { default-features = false, path = "src/co
 pangolin-runtime-system-params      = { default-features = false, path = "src/pallets/system/params" }
 pangoro-constants                   = { default-features = false, path = "../pangoro/src/constants" }
 pangoro-runtime-system-params       = { default-features = false, path = "../pangoro/src/pallets/system/params" }
+to-substrate-backing                = { default-features = false, path = "../../../frame/wormhole/backing/crab" }
```
HackFisher, line 60: The old crab backing is not the substrate backing we need.

Actually, old crab backing will be deleted after https://github.com/darwinia-network/darwinia/issues/622 being fix, the storage value in it will be migrated in to the mapping token contract of RING, as a value of account (RING Pallet Id's to dvm account)  cc @AsceticBear 

For Pangolin, I think we can just remove this pallet, and to-substrate-backing is used for connecting Pangoro.

Please reach me if you have questions.

AurevoirXavier: I just delete it, but not push yet.

Okay, I'll add a migration move `Crab*` to `Pangoro*`

HackFisher: We can just delete the old crab backing pallet in the runtime, and also delete the codes of the pallet.

For next runtime upgrade of Darwinia/Crab, we can replace the Crab Backing(in Darwinia side) with an to-substate-backing pallet(but use the same pallet id, because the account of the pallet id is holding the RING balances).

For Crab, the storage `TotalMappedRing` Crab Issuing (Crab side) and the Crab Issuing pallet itself should be deleted, the value in the storage `TotalMappedRing` can be migrated to the Mapping Token Contract of RING. (if the contract is not ready, we can document it in a github issue& email temporally and migrate later.)

 for the migration

HackFisher (approved):

AurevoirXavier:
Companion: https://github.com/darwinia-network/darwinia/pull/723

PR #789, opened 2021-08-31
confirm deliver result from remote

xiaoch05:


HackFisher:
Need to merge master due to project structure changes: #782

frame/evm/precompile/contracts/encoder/src/lib.rs
```diff
@@ -18,17 +18,29 @@
 
 #![cfg_attr(not(feature = "std"), no_std)]
 
+extern crate alloc;
```
AurevoirXavier, line 21: For substrate pallet we use `use sp_std::prelude::*` generally.

frame/wormhole/issuing/ethereum/src/lib.rs
```diff
@@ -292,29 +292,39 @@ pub mod pallet {
...
+				TokenRegisterInfo::decode(&input[8..]).map_err(|_| Error::<T>::InvalidInputData)?;
+			Self::finish_token_registered(register_info.0, register_info.1, register_info.2);
+			Ok(().into())
+		}
+
+		// when user burn their mapped tokens to unlock remote origin token, contract will deliver
+		// the burn information by this method
+		#[pallet::weight(<T as Config>::WeightInfo::burn_and_remote_unlock())]
+		#[transactional]
+		pub fn burn_and_remote_unlock(
```
HackFisher, line 305: The name is misleading, and the dispatch call only do `deposit_burn_token_event` ?

xiaoch05: this call will deposit the burn token event and trigger schedule_mmr_root to request authorities to sign mmr root, and then relay this event to ethereum chain.

frame/evm/precompile/contracts/encoder/src/lib.rs
```diff
@@ -23,12 +23,22 @@ use core::marker::PhantomData;
...
+use darwinia_support::s2s::RelayMessageCaller;
 use dp_evm::Precompile;
 // --- paritytech ---
 use frame_support::dispatch::{Dispatchable, GetDispatchInfo, PostDispatchInfo};
+use sp_std::prelude::Vec;
 
+const PALLET_DIG_LEN: usize = 4;
+const METHOD_DIG_LEN: usize = 4;
+const ACTION_LEN: usize = PALLET_DIG_LEN + METHOD_DIG_LEN;
+const BURN_AND_REMOTE_UNLOCK_METHOD: &[u8] = b"burn_and_remote_unlock()";
```
AurevoirXavier, line 37: Are these brackets necessary?

frame/wormhole/issuing/ethereum/src/lib.rs
```diff
@@ -292,29 +292,39 @@ pub mod pallet {
...
-				return Err(Error::<T>::UnsupportedAction.into());
-			}
+			let register_info =
+				TokenRegisterInfo::decode(&input[8..]).map_err(|_| Error::<T>::InvalidInputData)?;
+			Self::finish_token_registered(register_info.0, register_info.1, register_info.2);
+			Ok(().into())
+		}
+
+		// when user burn their mapped tokens to unlock remote origin token, contract will deliver
+		// the burn information by this method
```
HackFisher, line 301:
```suggestion
		// When user burn their mapped tokens to unlock remote origin token, mapping token factory will use precompile to call this, this call will deposit the burn token event and trigger schedule_mmr_root to request authorities to sign mmr root, and then relay this event to ethereum chain.
```

frame/wormhole/issuing/ethereum/src/lib.rs
HackFisher:
```suggestion
		pub fn deposit_burn_token_event_from_precompile(
```

frame/evm/precompile/contracts/encoder/src/lib.rs
```diff
@@ -47,29 +57,93 @@ where
...
+		if input.len() < ACTION_LEN {
 			return Err(ExitError::Other("input length less than 4 bytes".into()));
 		}
-		let selector = &input[0..SELECTOR];
-		let call: T::Call = match selector {
-			_ if selector == <from_substrate_issuing::Pallet<T>>::digest() => {
-				from_substrate_issuing::Call::<T>::asset_burn_event_handle(input.to_vec()).into()
+		let pallet_digest = &input[0..PALLET_DIG_LEN];
+		let method_digest = &input[PALLET_DIG_LEN..ACTION_LEN];
+		let output = match pallet_digest {
```
AurevoirXavier, line 65: An unnecessary match I think.

frame/evm/precompile/contracts/encoder/src/lib.rs
xiaoch05: Yes this can be deleted if we define it the same as in the contract. Here we use this format to describe it as a function. In fact, we should also add the parameters, but for simplicity, we ignore them.

frame/wormhole/backing/s2s/src/lib.rs
```diff
@@ -102,6 +103,8 @@ pub mod pallet {
 		TokenLocked(Token, AccountId<T>, EthereumAddress, U256),
 		/// Token unlocked [token, recipient, value]
 		TokenUnlocked(Token, AccountId<T>, U256),
+		/// Token locked confirmed from remote [token, user, result]
+		TokenLockedConfirmed(Token, AccountId<T>, bool),
```
AurevoirXavier, line 107: I think this line doesn't affect anything.

```
#[pallet::metadata(T::AccountId = "AccountId")]
```

frame/support/src/lib.rs
```diff
@@ -83,9 +90,21 @@ pub mod s2s {
...
+	pub fn nonce_to_message_id(lane_id: &[u8], nonce: u64) -> BridgeMessageId {
+		let mut message_id: BridgeMessageId = Default::default();
+		message_id[4..8].copy_from_slice(&lane_id[..4]);
+		message_id[8..].copy_from_slice(&nonce.to_be_bytes());
+		message_id
+	}
 }
 
 pub type PalletDigest = [u8; 4];
+pub type BridgeMessageId = [u8; 16];
```
AurevoirXavier, line 107: Docs require.

```
/// [0..4]  bytes ---- reserved
/// [4..8]  bytes ---- laneID
/// [8..16] bytes ---- message nonce
```

frame/support/src/lib.rs
```diff
@@ -83,9 +90,21 @@ pub mod s2s {
 		result.copy_from_slice(&encoded);
 		result
 	}
+
+	// 128 bit or 16 bytes
+	// [0..4]  bytes ---- reserved
+	// [4..8]  bytes ---- laneID
+	// [8..16] bytes ---- message nonce
```
AurevoirXavier, line 97: Could move this doc below.

frame/wormhole/issuing/s2s/src/lib.rs
```diff
@@ -117,28 +117,23 @@ pub mod pallet {
...
-				log::trace!("No action match this input selector");
+			let fee = Self::transform_dvm_balance(burn_info.fee);
+			if let Some(fee_account) = T::FeeAccount::get() {
+				// Since fee account will represent use to make a cross chain call, give fee to fee account here.
+				// the fee transfer path
+				// user -> mapping_token_factory(caller) -> fee_account -> fee_fund -> relayers
+				<T as Config>::RingCurrency::transfer(&caller, &fee_account, fee, KeepAlive)?;
 			}
+
+			Self::burn_and_remote_unlock(fee, burn_info)?;
```
HackFisher, line 136: https://github.com/darwinia-network/darwinia-common/issues/816

Need refactoring, may be in another PR.

HackFisher:
Companion PR for review together: https://github.com/darwinia-network/darwinia-bridge-sol/pull/66

frame/wormhole/backing/s2s/src/lib.rs
```diff
@@ -122,8 +125,15 @@ pub mod pallet {
...
 		/// send relay message failed
 		SendMessageFailed,
+		/// message nonce dumplicated
+		NonceDumplicated,
 	}
 
+	#[pallet::storage]
+	#[pallet::getter(fn locked_queue)]
+	pub type LockedQueue<T: Config> =
+		StorageMap<_, Blake2_128Concat, [u8; 16], (AccountId<T>, Token), ValueQuery>;
```
AurevoirXavier, line 135: `Blake2_128Concat` could replace with `Identity`.

`[u8; 16]` could replace with `BridgeMessageId`.

AurevoirXavier:
Empty hooks block can be remove after #805

frame/evm/precompile/contracts/encoder/src/lib.rs
```diff
@@ -47,29 +57,93 @@ where
...
+			_ if pallet_digest == <from_substrate_issuing::Pallet<T>>::digest() => {
+				match method_digest {
+					_ if method_digest
+						== &sha3::Keccak256::digest(BURN_AND_REMOTE_UNLOCK_METHOD)
+							[..METHOD_DIG_LEN] =>
```
AsceticBear, line 70: Can we do like the following:

```rs
let digest = |method| { sha3::Keccak256::digest(method) }[..METHOD_DIG_LEN];

_ if method_digest == digest(BURN_AND_REMOTE_UNLOCK_METHOD) => {...}
```

frame/evm/precompile/contracts/encoder/src/lib.rs
hujw77: Is the method selector from the contract, if so the parameters should not be omitted

frame/evm/precompile/contracts/encoder/src/lib.rs
xiaoch05: I have tried before, but it reports an unknown size error. Any idea?

frame/evm/precompile/contracts/encoder/src/lib.rs
xiaoch05: This selector is used to map the caller. And the method is not the real function of the contract.

frame/evm/precompile/contracts/encoder/src/lib.rs
AurevoirXavier: I don't know the context.

But basically, there are two ways.

1. reference the return type
2. specify the return type, IMO the `fn digest` could return a generic

AsceticBear: I have tried this way in my local, it works. FYI

```rs
let digest = |param_method, expected_method| -> bool {
	&sha3::Keccak256::digest(param_method)[..METHOD_DIG_LEN] == expected_method
};

_ if digest(method_digest, BURN_AND_REMOTE_UNLOCK_METHOD) => {..}
```

AurevoirXavier: If that then rename to `check_digest`

AsceticBear: Right.

xiaoch05: This way looks good.

xiaoch05:
I test failed at the last time since some lastest modification, please don't merge until I confirmed the issue.
Why I can't change it to draft now?

frame/wormhole/issuing/s2s/src/lib.rs
xiaoch05: Leave this to another PR.

xiaoch05:
local test passed. Please help to review again.

frame/evm/precompile/contracts/encoder/src/lib.rs
AurevoirXavier: Simply use if-else here would be more concise?

frame/evm/precompile/contracts/encoder/src/lib.rs
```diff
@@ -47,30 +57,82 @@ where
...
+						return Err(ExitError::Other(
+							"No such method in pallet substrate issuing".into(),
+						));
+					}
+				}
 			}
-			_ if selector == <from_ethereum_issuing::Pallet<T>>::digest() => {
-				from_ethereum_issuing::Call::<T>::mapping_factory_event_handle(input.to_vec())
-					.into()
+			_ if pallet_digest == <from_ethereum_issuing::Pallet<T>>::digest() => {
```
AurevoirXavier, line 91: Also here

frame/wormhole/issuing/ethereum/src/lib.rs
```diff
@@ -292,29 +291,41 @@ pub mod pallet {
...
+			origin: OriginFor<T>,
+			input: Vec<u8>,
+		) -> DispatchResultWithPostInfo {
+			let caller = ensure_signed(origin)?;
+			ensure!(input.len() >= 8, <Error<T>>::InvalidInput);
+			let factory = MappingFactoryAddress::<T>::get();
+			let factory_id = <T as darwinia_evm::Config>::AddressMapping::into_account_id(factory);
+			ensure!(factory_id == caller, <Error<T>>::NoAuthority);
+			let burn_info =
+				TokenBurnInfo::decode(&input[8..]).map_err(|_| Error::<T>::InvalidInputData)?;
```
AurevoirXavier, line 316: Could you document what 8 for?

frame/evm/precompile/contracts/encoder/src/lib.rs
xiaoch05: maybe we will have other pallets using this encoder precompile? Not only s2s and ethereum pallets.

frame/wormhole/issuing/ethereum/src/lib.rs
xiaoch05: It is better to move it to decoder precompile.

HackFisher (approved):
Agree that this PR require further cleanup, but also good to go if put next changes in the PR to resolve #816 considering that we need a running test bridge for wormhole ui dev.

Need more cleanup for better naming and simplicity.

frame/wormhole/backing/s2s/src/lib.rs
```diff
@@ -267,6 +283,43 @@ pub mod pallet {
...
+					let _ = T::RingCurrency::transfer(
+						&Self::pallet_account_id(),
+						&user,
+						value.low_u128().unique_saturated_into(),
+						AllowDeath,
+					);
+				}
+			}
+			Self::deposit_event(Event::TokenLockedConfirmed(token, user, result));
+			return 1;
```
AsceticBear, line 320: We should not return a const weight number, better to record the read and write times in this function. A example like this:

https://github.com/paritytech/parity-bridges-common/blob/master/modules/token-swap/src/lib.rs#L484

frame/support/src/lib.rs
AsceticBear: Why not keeping the message-id format consistent with parity-bridges-common, `(lane_id, message nonce)`. What is the purpose of the reserved bytes?

frame/evm/precompile/contracts/encoder/src/lib.rs
```diff
@@ -47,30 +57,82 @@ where
...
-			cost: 0,
-			output: call.encode(),
+			cost: 20000,
+			output,
 			logs: Default::default(),
 		})
 	}
 }
+
+impl<T> DispatchCallEncoder<T> {
```
AsceticBear, line 134: Actually, we do not need this `impl`, `match_digest` has no relationship with `DispatchCallEncoder`, right?

frame/support/src/lib.rs
xiaoch05: Maybe this definition is not so good. But we should not use `(lane_id, message nonce)` directly as in parity-bridges-common. We need a common definition of `message id`. And this `message id` should be generic for sub<>sub， ethereum<>darwina， bsc<>darwinia, etc.
The reserved is not used now. Maybe it will be filled by chainID or some other identities.

frame/evm/precompile/contracts/encoder/src/lib.rs
xiaoch05: Yes, a normal function also works.

frame/wormhole/backing/s2s/src/lib.rs
xiaoch05: Here, we use a const low number to ensure the caller not failed.
Furthermore, we need a more accurate calculation of this weight. Because the caller restricts this weight. 
Firstly we should not do so many things in the callback.
Secondly, we need to give an accurate weight of this callback.

AsceticBear: Yeap, there are some weight limits for a callback function, we can create an issue about this and fix it later.

HackFisher:
After checking with @xiaoch05 , he would like to address unresolved conversations in next PRs, related issues: #819 #816 .

Merging this for the requirement of wormhole ui dev.

AurevoirXavier:
Would you mind adding some info to describe this PR? @xiaoch05

PR #791, opened 2021-09-01
Adjust Structure

AurevoirXavier:


node/cli/Cargo.toml
```diff
@@ -21,7 +21,7 @@ structopt = { version = "0.3.21", optional = true }
 # darwinia-network
 bsc-primitives = { path = "../../primitives/bsc" }
 dp-consensus   = { path = "../../primitives/consensus" }
-drml-service   = { path = "../service" }
+service        = { package = "drml-service", default-features = false, path = "../service" }
```
AsceticBear, line 24: `default-features = false` is necessnary here?

AurevoirXavier: Yes. Otherwise, we can not control its features.

AsceticBear: Weird, why the old version can compile without `default-features = false`?

AurevoirXavier: Because all the features set to enable by default.

AsceticBear (approved):
LGTM

PR #802, opened 2021-09-03
Usage document

fewensa:
A simple document to start pangolin and pangoro chain


AurevoirXavier:
Recommend to put this on README and provide a script to deploy the test env.

PR #817, opened 2021-09-14
Update dvm from 426 to 435

AsceticBear:
Companion pull request: https://github.com/darwinia-network/darwinia/pull/738


frame/dvm-dynamic-fee/src/lib.rs
```diff
@@ -81,12 +81,16 @@ pub mod pallet {
...
 		/// Set the target gas price.
-		#[pallet::weight(T::DbWeight::get().writes(1))]
-		fn note_min_gas_price_target(
+		#[pallet::weight((T::DbWeight::get().writes(1), DispatchClass::Mandatory))]
+		pub fn note_min_gas_price_target(
 			origin: OriginFor<T>,
 			target: U256,
 		) -> DispatchResultWithPostInfo {
 			ensure_none(origin)?;
+			assert!(
```
HackFisher, line 90: `assert!` looks dangerous, will it cause the runtime panic?

tests/dvm/tests/0_test-rpc-block.js
```diff
@@ -20,7 +20,7 @@ describe("Test Block RPC", function () {
 				"0x00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000",
 			miner: "0x0000000000000000000000000000000000000000",
 			number: 0,
-			receiptsRoot: "0x1dcc4de8dec75d7aab85b567b6ccd41ad312451b948a7413f0a142fd40d49347",
+			receiptsRoot: "0x56e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421",
```
HackFisher, line 23: why the receiptsRoot chagnes?

HackFisher (changes requested):

frame/dvm-dynamic-fee/src/lib.rs
AsceticBear: No. We use `<TargetMinGasPrice<T>>::take()` in `on_finalize`. 

1. More discuss here https://github.com/paritytech/frontier/pull/435#discussion_r680966368
2. Substrate timestamp usage: https://github.com/paritytech/substrate/blob/master/frame/timestamp/src/lib.rs#L196

tests/dvm/tests/0_test-rpc-block.js
AsceticBear: The old way of updating the receipt_root is not right. check https://github.com/paritytech/frontier/pull/426

frame/dvm-dynamic-fee/src/lib.rs
HackFisher: What will happen to the block producers if they intent to insert multiple target gas?

Especially when do runtime upgrade, on_runtime_upgrade is after last runtime's final block's finalize and before next runtime's first block, all block producers might do same thing to note, if all of them are stalled, the chain will be stalled.

I worry that this is some indentical behavior for all clients.

Timestamp pallet do not happen might just because there were no runtime upgrade including such changes in production network.

We need further learn of client behavior or testing of this.

HackFisher: In case we forget, we need to do migration to clean/remove `<TargetMinGasPrice<T>>` for runtime upgrade.

HackFisher (approved):
LGTM

AurevoirXavier (approved):

Issue #810, opened 2021-09-09
The internal transaction's error information is not enough


xiaoch05:
when we send an internal transaction to DVM and it failed. Then we get such error information. It has nothing about the reason for the failed execution.
```
ethereum.FailedInternalTx
The internal transaction failed
```

PR #837, opened 2021-09-23
Update internal transaction dispatch error

AsceticBear:
Enrich the internal transaction dispatch error type.

### DIFF
--- a/frame/dvm/src/lib.rs
+++ b/frame/dvm/src/lib.rs
@@ -185,7 +185,9 @@ pub mod pallet {
 		/// Pre-log is present, therefore transact is not allowed.
 		PreLogExists,
 		/// The internal transaction failed.
-		FailedInternalTx,
+		InternalTransactionExitError,
+		InternalTransactionRevertError,
+		InternalTransactionFatalError,
 		/// The internal call failed.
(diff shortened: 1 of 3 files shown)


HackFisher:
This fixes #810 ?

frame/dvm/src/lib.rs
```diff
@@ -620,12 +621,11 @@ impl<T: Config> InternalTransactHandler for Pallet<T> {
 		match info {
 			CallOrCreateInfo::Call(info) => match info.exit_reason {
 				ExitReason::Succeed(_) => Ok(info.value),
-				_ => {
-					log::error!("Executing internal transaction error happened");
-					Err(Error::<T>::FailedInternalTx.into())
-				}
+				ExitReason::Error(_) => Err(<Error<T>>::InternalTransactionExitError.into()),
+				ExitReason::Revert(_) => Err(<Error<T>>::InternalTransactionRevertError.into()),
```
xiaoch05, line 625: does this only show the error type `revert`? Can we catch the detail of the error?

HackFisher (approved):

xiaoch05 (approved):

PR #844, opened 2021-09-26
Update the mapped token name

xiaoch05:


frame/wormhole/issuing/ethereum/src/lib.rs
```diff
@@ -86,6 +86,7 @@ pub mod pallet {
 		type EcdsaAuthorities: RelayAuthorityProtocol<Self::BlockNumber, Signer = EthereumAddress>;
 		type WeightInfo: WeightInfo;
 		type InternalTransactHandler: InternalTransactHandler;
+		type BridgeChainName: Get<[u8; 32]>;
```
HackFisher, line 89:
```suggestion
		type BackingChainName: Get<[u8; 32]>;
```

frame/wormhole/issuing/ethereum/src/lib.rs
```diff
@@ -238,6 +239,7 @@ pub mod pallet {
 				register_info.decimals.as_u32() as u8,
 				backing_address,
```
HackFisher, line 239: What is backing address and source address?

frame/wormhole/issuing/ethereum/src/lib.rs
```diff
@@ -238,6 +239,7 @@ pub mod pallet {
...
 				register_info.token_address,
```
HackFisher, line 240: Found it is called as source in smart contract?

In the token protocol context, we'd better to unify and using backing and issuing, because source and target are used for message layer.

HackFisher (commented):
Request better naming

HackFisher (approved):

node/runtime/pangolin/src/pallets/from_substrate_issuing_.rs
```diff
@@ -89,6 +90,7 @@ impl ToEthAddress<AccountId32> for TruncateToEthAddress {
 frame_support::parameter_types! {
 	pub const S2sIssuingPalletId: PalletId = PalletId(*b"da/s2sis");
 	pub const PangoroChainId: ChainId = PANGORO_CHAIN_ID;
+	pub PangoroName: [u8; 32] = to_bytes32(b"Pangoro").into();
```
AsceticBear, line 93: Recommend to give an alas to `[u8; 32]`, to avoid using it everywhere.

xiaoch05:
We have another problem when using this format of the token name.
In order to be compatible with some erc20 token names defined by bytes32, we restrict the name to 32 bytes.
If we add this `(BackingChain>IssuingChain)` to the end of the original name, it may become too long.

HackFisher:
> We have another problem when using this format of the token name.
> In order to be compatible with some erc20 token names defined by bytes32, we restrict the name to 32 bytes.
> If we add this `(BackingChain>IssuingChain)` to the end of the original name, it may become too long. @HackFisher

Since names are all normalized to the string name in mapping token contract:

https://github.com/darwinia-network/darwinia-bridge-sol/blob/bd3313a3f12b9118a2e1ae90bf58af83776e085d/contracts/mapping-token/contracts/darwinia/MappingERC20.sol#L21

Will the problem you mentioned still hold?

frame/wormhole/issuing/ethereum/src/lib.rs
xiaoch05: > > We have another problem when using this format of the token name.
> > In order to be compatible with some erc20 token names defined by bytes32, we restrict the name to 32 bytes.
> > If we add this `(BackingChain>IssuingChain)` to the end of the original name, it may become too long. @HackFisher
> 
> Since names are all normalized to the string name in mapping token contract:
> 
> https://github.com/darwinia-network/darwinia-bridge-sol/blob/bd3313a3f12b9118a2e1ae90bf58af83776e085d/contracts/mapping-token/contracts/darwinia/MappingERC20.sol#L21
> 
> Will the problem you mentioned still hold?

The contract can save this long type. And I'll remove this restrict https://github.com/darwinia-network/darwinia-common/blob/master/primitives/asset/src/token.rs#L35, this can be fixed.
But should we restrict the mapped times to prevent this name becomes long and long?

xiaoch05:
> > We have another problem when using this format of the token name.
> > In order to be compatible with some erc20 token names defined by bytes32, we restrict the name to 32 bytes.
> > If we add this `(BackingChain>IssuingChain)` to the end of the original name, it may become too long. @HackFisher
> 
> Since names are all normalized to the string name in mapping token contract:
> 
> https://github.com/darwinia-network/darwinia-bridge-sol/blob/bd3313a3f12b9118a2e1ae90bf58af83776e085d/contracts/mapping-token/contracts/darwinia/MappingERC20.sol#L21
> 
> Will the problem you mentioned still hold?

The contract can save this long type. And I'll remove this restrict https://github.com/darwinia-network/darwinia-common/blob/master/primitives/asset/src/token.rs#L35, this can be fixed.
But should we restrict the mapped times to prevent this name becomes long and long?

frame/wormhole/issuing/s2s/src/lib.rs
```diff
@@ -170,18 +170,20 @@ pub mod pallet {
 						&name,
 						&symbol,
 						option.decimal,
-						backing,
+						backing_address,
 						token_info.address,
```
HackFisher, line 174: Could we generate new symbol and name here? We will require less parameters. and we do not have store prefix and pass backing_chain_name in contract template.

xiaoch05: Yes we can do this.

HackFisher:
> We have another problem when using this format of the token name.
> In order to be compatible with some erc20 token names defined by bytes32, we restrict the name to 32 bytes.
> If we add this `(BackingChain>IssuingChain)` to the end of the original name, it may become too long. @HackFisher

Another proposal for name could be as following, since the mapping token is located at the issuing chain, and easy to aware of that:

name postfix: "[{backing_chain_short_name}>"

Then for example, RING from Darwinia to Crab will be "Darwinia Network Native Token[Darwinia>", then to Moonriver will be "Darwinia Network Native Token[Darwinia>[Crab>"

xiaoch05:
> > We have another problem when using this format of the token name.
> > In order to be compatible with some erc20 token names defined by bytes32, we restrict the name to 32 bytes.
> > If we add this `(BackingChain>IssuingChain)` to the end of the original name, it may become too long. @HackFisher
> 
> Another proposal for name could be as following, since the mapping token is located at the issuing chain, and easy to aware of that:
> 
> name postfix: "[${backing_chain_short_name}>"
> 
> Then for example, RING from Darwinia to Crab will be "Darwinia Network Native Token[Darwinia>", then to Moonriver will be "Darwinia Network Native Token[Darwinia>[Crab>"

I agreen this concise format. Users only need to be told where the token comes from.

HackFisher:
Please update according to latest documents: https://github.com/darwinia-network/darwinia-bridge-sol/blob/master/docs/Wormhole%20Mapping%20Token%20Metadata%20Standards.md

HackFisher:
@xiaoch05 compile failed.

HackFisher (approved):

HackFisher:
@xiaoch05 still some compile failure in tests.

xiaoch05:
The latest test result.
image

PR #847, opened 2021-09-29
s2s: move redeem steps into contract

xiaoch05:


HackFisher:
@xiaoch05 Any update on this?

PR #870, opened 2021-10-18
Add message fee verification with market fee

### LINKED ISSUE #862: Check message fee from the fee-market by customizing LaneMessageVerifier
https://github.com/darwinia-network/darwinia-common/blob/2162174d7c370f5f587954240b6ccd7497c94a2c/node/runtime/pangoro/src/pallets/bridge_message.rs#L60

boundless-forest:
Close #862, #863

~In the process of solving this problem, I find that adding the fee verification in `FeeMessageDeliveryAndDispatchPayment` would be much reasonable in our implementation. We don't need to hack a `DarwiniaLaneMessageVerifier` in this way.~

Just follow the way described in the issue, make a hacked `DarwiniaLaneMessageVerifier`.


HackFisher:
Could we customize and impl a new LaneMessageVerifier in our runtime?

No need to hack this way as my understanding.

PR #887, opened 2021-10-25
Make s2s runtime call encode better

AsceticBear:
Simplify the original runtime call encode process by introducing `PayloadCreate` and `CallParams`.

The default `encode_call` accepts the pallet index and runtime call params to generate call binay, then `payload` gives the final result out.

```rs
/// Creating a concrete message payload which would be relay to target chain.
pub trait PayloadCreate<AccountId, MessagePayload>
where
	AccountId: Encode + Decode,
{
	fn encode_call(
		pallet_index: u8,
		call_params: CallParams<AccountId>,
	) -> Result<Vec<u8>, &'static str> {
		let mut encoded = vec![pallet_index];
		encoded.append(&mut call_params.encode());
		Ok(encoded)
	}

	fn payload(
		spec_version: u32,
		weight: u64,
		call_params: CallParams<AccountId>,
	) -> Result<MessagePayload, &'static str>;
}
```

A simple codec encode example to help the reviewing:

```rs
use codec::{Decode, Encode};

#[derive(Debug, PartialEq, Encode, Decode)]
enum Runtime {
	#[codec(index = 20)]
	Sub2SubBacking(S2SBackingCall),
}

#[derive(Debug, PartialEq, Encode, Decode, Clone)]
pub enum S2SBackingCall {
	#[codec(index = 2)]
	unlock_from_remote(bool, u64, MockData),
}

#[derive(Debug, PartialEq, Encode, Decode, Clone)]
pub enum MockData {
	#[codec(index = 5)]
	Data(bool)
}

fn main() {
	let call = S2SBackingCall::unlock_from_remote(true, 2, MockData::Data(true));
	assert_eq!(call.encode(), vec![2, 1, 2, 0, 0, 0, 0, 0, 0, 0]);
	let runtime = Runtime::Sub2SubBacking(call.clone());
	assert_eq!(runtime.encode(), vec![20, 2, 1, 2, 0, 0, 0, 0, 0, 0, 0]);
}
```


primitives/s2s/src/lib.rs
```diff
@@ -0,0 +1,168 @@
...
+
+	fn payload(
+		spec_version: u32,
+		weight: u64,
+		call_params: CallParams<AccountId>,
+	) -> Result<MessagePayload, &'static str>;
+}
+
+#[cfg(test)]
+mod test {
```
AsceticBear, line 62: Some tests to ensure the call encoded result the same after this refactor.

primitives/s2s/src/lib.rs
```diff
@@ -0,0 +1,168 @@
...
+/// The parameters box for the pallet runtime call.
+#[derive(Encode, Decode, Debug, PartialEq, Eq, Clone)]
+pub enum CallParams<AccountId> {
+	#[codec(index = 0)]
+	S2sIssuingPalletRegisterFromRemote(Token),
+	#[codec(index = 1)]
+	S2sIssuingPalletIssueFromRemote(Token, H160),
+	#[codec(index = 2)]
+	S2sBackingPalletUnlockFromRemote(AccountId, S2sRemoteUnlockInfo),
+}
```
xiaoch05, line 38: we must add the prefix XXXPallet to indicate which pallet this remote call belongs. It seems not good.

primitives/s2s/src/lib.rs
```diff
@@ -0,0 +1,168 @@
...
+	}
+
+	fn payload(
+		spec_version: u32,
+		weight: u64,
+		call_params: CallParams<AccountId>,
+	) -> Result<MessagePayload, &'static str>;
+}
+
+#[cfg(test)]
```
xiaoch05, line 61: put the test to a single file

primitives/contract/src/mapping_token_factory/s2s.rs
```diff
@@ -95,7 +96,7 @@ impl S2sRemoteUnlockInfo {
 		])
 	}
 
-	pub fn decode(data: &[u8]) -> AbiResult<Self> {
+	pub fn eth_decode(data: &[u8]) -> AbiResult<Self> {
```
xiaoch05, line 99:
```suggestion
	pub fn abi_decode(data: &[u8]) -> AbiResult<Self> {
```

node/runtime/pangolin/src/pallets/from_substrate_issuing_.rs
```diff
@@ -5,55 +5,37 @@ use sp_runtime::AccountId32;
...
+const PANGORO_S2S_BACKING_PALLET_INDEX: u8 = 20;
+/// Create message payload according to the call parameters.
+pub struct PangoroPayloadCreator;
+impl PayloadCreate<AccountId, ToPangoroMessagePayload> for PangoroPayloadCreator {
+	fn payload(
+		spec_version: u32,
+		weight: u64,
+		call_params: CallParams<AccountId>,
+	) -> Result<ToPangoroMessagePayload, &'static str> {
+		let (submitter, call) = match call_params.clone() {
```
xiaoch05, line 21: still, I think this match is complex. and the type CallParams is not intuitive.

primitives/s2s/src/lib.rs
```diff
@@ -0,0 +1,168 @@
...
+pub trait PayloadCreate<AccountId, MessagePayload>
+where
+	AccountId: Encode + Decode,
+{
+	fn encode_call(
+		pallet_index: u8,
+		call_params: CallParams<AccountId>,
+	) -> Result<Vec<u8>, &'static str> {
+		let mut encoded = vec![pallet_index];
+		encoded.append(&mut call_params.encode());
```
xiaoch05, line 50: this method to encode the remote call has a problem that we can't know exactly which method to encode. Scale or others? And we know we use the scale encode. But if the implementation of the scale encode changed(version upgrade), then it can't update with it.

xiaoch05 (changes requested):

primitives/s2s/src/lib.rs
AsceticBear: I already added `S2sBackingPallet` or `S2sIssuingingPallet`. Or i am missing something?

node/runtime/pangolin/src/pallets/from_substrate_issuing_.rs
AsceticBear: After s2s-audit is merged, the payload() signature has a submitter param, this match can be deleted.

primitives/s2s/src/lib.rs
HackFisher: What if two call from two separate pallet have the same call index?

AsceticBear: Good question. Just give the actual index. No conflict, a demo like this:

```rs
use codec::Encode;

#[derive(Encode)]
enum T {
	#[codec(index = 1)]
	A,
	#[codec(index = 1)]
	B,
}
fn main() {
	assert_eq!(T::A.encode(), vec![1]);
	assert_eq!(T::B.encode(), vec![1]);
}
```

node/runtime/pangolin/src/pallets/from_substrate_issuing_.rs
```diff
@@ -6,55 +6,33 @@ use sp_runtime::AccountId32;
...
+		spec_version: u32,
+		weight: u64,
+		call_params: CallParams,
+	) -> Result<ToPangoroMessagePayload, &'static str> {
+		let call = Self::encode_call(PANGORO_S2S_BACKING_PALLET_INDEX, call_params)?;
+
+		Ok(ToPangoroMessagePayload {
+			spec_version,
+			weight,
+			origin: bp_message_dispatch::CallOrigin::SourceAccount(submitter),
```
xiaoch05, line 28: how can we use this payload func when we want to send message by SourceRoot, not this SourceAccount?

xiaoch05 (changes requested):

node/runtime/pangoro/src/pallets/to_substrate_backing_.rs
```diff
@@ -6,84 +6,32 @@ use bp_runtime::{messages::DispatchFeePayment, ChainId};
...
-/// You must update the index here if you change the call order in Pangolin runtime.
-#[derive(Encode, Decode, Debug, PartialEq, Eq, Clone)]
-#[allow(non_camel_case_types)]
-pub enum PangolinSub2SubIssuingCall {
-	#[codec(index = 0)]
-	register_from_remote(TokenMetadata),
-	#[codec(index = 1)]
-	issue_from_remote(H160, U256, H160),
-}
+const PANGOLIN_S2S_ISSUING_PALLET_INDEX: u8 = 49;
```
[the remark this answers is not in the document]
AsceticBear: We can indeed. Now, this index is only used in specific pallet runtime config, such as using `PANGORO_S2S_BACKING_PALLET_INDEX` in `from-substrate-issuing` and `PANGOLIN_S2S_ISSUING_PALLET_INDEX` in `to-substrate-backing`. I think we can together with them in a common runtime primitive mod if we have more such indexes in the future. wdyt?

primitives/contract/src/mapping_token_factory/s2s.rs
```diff
@@ -67,18 +69,18 @@ impl Sub2SubMappingTokenFactory {
...
-#[derive(Debug, PartialEq, Eq)]
+#[derive(Encode, Decode, Debug, PartialEq, Eq, Clone)]
 pub struct S2sRemoteUnlockInfo {
 	pub spec_version: u32,
 	pub weight: u64,
 	pub token_type: u32,
 	pub original_token: H160,
 	pub amount: U256,
-	pub recipient: Vec<u8>,
+	pub recipient: AccountId32,
```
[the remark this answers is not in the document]
AsceticBear: O, In what circumstances would the `recipient` be the H160?  The `unlock_from_remote` call signature is:

```rs
pub fn unlock_from_remote(
    origin: OriginFor<T>,
    token_address: H160,
    amount: U256,
    recipient: AccountId<T>,   // AccountId now
) -> DispatchResultWithPostInfo {}
```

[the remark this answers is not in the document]
xiaoch05: when we support the erc20 token, this recipient would be H160.

node/runtime/pangolin/src/pallets/from_substrate_issuing_.rs
```diff
@@ -6,55 +6,33 @@ use sp_runtime::AccountId32;
...
+			call,
+			dispatch_fee_payment: DispatchFeePayment::AtSourceChain,
```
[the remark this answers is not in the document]
AsceticBear: If you want to choose different `DispatchFeePayment`, one simple way is to pass `dispatch_fee_payment` as `payload` function parameter. Then, the callers can choose different `DispatchFeePayment` in the pallet. Do you think this is necessary for the current implementation?  IMO, the current way is enough if we do not need different dispatch fee payment in one pallet.

primitives/s2s/src/lib.rs
xiaoch05: it's a little confusing to put methods in different pallets together.

node/runtime/pangolin/src/pallets/from_substrate_issuing_.rs
AsceticBear: Two way:
1. Changing `origin: bp_message_dispatch::CallOrigin::SourceAccount(submitter),` to `origin: bp_message_dispatch::CallOrigin::SourceRoot`
2. Make call origin as a function param.

node/runtime/pangoro/src/pallets/to_substrate_backing_.rs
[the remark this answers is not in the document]
xiaoch05: I think these pallet indexes belong to the runtime of the target chain. They should be put together in the runtime of the source chain.

node/runtime/pangolin/src/pallets/from_substrate_issuing_.rs
xiaoch05: 1. Some funcs want to use SourceAccount while other funcs want to use RootAccount. They can not be described by a common payload() func
2. If we put it in param, we need introduce this type in pallet

node/runtime/pangolin/src/pallets/from_substrate_issuing_.rs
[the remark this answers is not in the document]
xiaoch05: The common `payload()` func causes hard code of these params I think. Every time when we want to change these, we must add them to the params.

node/runtime/pangolin/src/pallets/from_substrate_issuing_.rs
HackFisher: I think we should replace submitter with CallOrigin in the function parameter

node/runtime/pangolin/src/pallets/from_substrate_issuing_.rs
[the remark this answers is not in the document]
HackFisher: I think it might be better to remove `fn payload()` since it is no different with ToPangoroMessagePayload  constructor.

node/runtime/pangolin/src/pallets/from_substrate_issuing_.rs
AsceticBear: That‘s a nice idea, I will have a try.

frame/wormhole/backing/s2s/src/lib.rs
```diff
@@ -360,56 +359,63 @@ pub mod pallet {
...
+					T::MessageNoncer::inbound_latest_received_nonce(T::MessageLaneId::get());
+				let message_id = nonce_to_message_id(&T::MessageLaneId::get(), message_nonce);
+				Self::deposit_event(Event::TokenUnlocked(
+					message_id,
+					token_address,
+					recipient_id,
+					amount,
+				));
+			} else if recipient.len() == 20 {
+				// TODO: Reserved for ERC20 unlock
```
HackFisher, line 415: Return Invalid Error for now

AsceticBear:
@HackFisher @xiaoch05 @AurevoirXavier Ready for review. Some works about benchmarks are going on.

frame/wormhole/backing/s2s/src/lib.rs
AsceticBear: Fixed.

frame/wormhole/backing/s2s/src/lib.rs
```diff
@@ -107,10 +106,8 @@ pub mod pallet {
 		type BridgedChainId: Get<ChainId>;
 
 		/// Outbound payload used for s2s message
-		type OutboundPayload: Parameter + Size;
-
-		/// The call encoder to encode a remote dispatch call
-		type CallEncoder: EncodeCall<Self::AccountId, Self::OutboundPayload>;
+		type OutboundPayload: Parameter
```
HackFisher, line 109:
```suggestion
		type OutboundPayloadCreator: Parameter
```

frame/wormhole/issuing/s2s/src/lib.rs
```diff
@@ -92,12 +92,10 @@ pub mod pallet {
 		type ToEthAddressT: ToEthAddress<Self::AccountId>;
 
 		/// Outbound payload used for s2s message
-		type OutboundPayload: Parameter + Size;
+		type OutboundPayload: Parameter
```
HackFisher, line 95:
```suggestion
		type OutboundPayloadCreator: Parameter
```

frame/support/src/lib.rs
```diff
@@ -55,15 +54,11 @@ pub mod s2s {
...
-
 	pub trait ToEthAddress<A> {
 		fn into_ethereum_id(address: &A) -> H160;
 	}
 
-	// RelayMessageCaller send message to pallet-messages
+	// RelayMessageSender send message to pallet-messages
 	pub trait RelayMessageSender {
 		fn encode_send_message(
 			pallet_index: u32,
```
HackFisher, line 64:
```suggestion
			message_pallet_index: u32,
```

primitives/s2s/src/call_params.rs
```diff
@@ -0,0 +1,193 @@
...
+	#[codec(index = 1)]
+	S2sIssuingPalletIssueFromRemote(H160, U256, H160),
+	#[codec(index = 2)]
+	S2sBackingPalletUnlockFromRemote(H160, U256, Vec<u8>),
+}
+/// Creating a concrete message payload which would be relay to target chain.
+pub trait CreatePayload<SourceChainAccountId, TargetChainAccountPublic, TargetChainSignature>
+where
+	SourceChainAccountId: Encode,
+	Self: Sized,
```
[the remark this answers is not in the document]
HackFisher: Following is payload's type, I'm sure whether we can add more limitation to Payload associate type:
Aka. geting types <SourceChainAccountId, TargetChainAccountPublic, TargetChainSignature> from MessagePayload

```
/// Message payload type used by dispatch module.
#[derive(RuntimeDebug, Encode, Decode, Clone, PartialEq, Eq)]
pub struct MessagePayload<
	SourceChainAccountId,
	TargetChainAccountPublic,
	TargetChainSignature,
	Call,
> {
	/// Runtime specification version. We only dispatch messages that have the same
	/// runtime version. Otherwise we risk to misinterpret encoded calls.
	pub spec_version: SpecVersion,
	/// Weight of the call, declared by the message sender. If it is less than actual
	/// static weight, the call is not dispatched.
	pub weight: Weight,
	/// Call origin to be used during dispatch.
	pub origin: CallOrigin<SourceChainAccountId, TargetChainAccountPublic, TargetChainSignature>,
	/// Where the fee for dispatching message is paid?
	pub dispatch_fee_payment: DispatchFeePayment,
	/// The call itself.
	pub call: Call,
}
```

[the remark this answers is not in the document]
xiaoch05: Self: Sized this line restricts the type CreatePayload itself. Not its inner type Payload. Is it necessary?

primitives/s2s/src/call_params.rs
```diff
@@ -0,0 +1,193 @@
...
+	fn create(
+		origin: CallOrigin<SourceChainAccountId, TargetChainAccountPublic, TargetChainSignature>,
+		spec_version: u32,
+		weight: u64,
+		call_params: CallParams,
+		dispatch_fee_payment: DispatchFeePayment,
+	) -> Result<Self::Payload, &'static str>;
+}
+
+impl CreatePayload<AccountId32, MultiSigner, MultiSignature> for () {
```
xiaoch05, line 64: do we need to implement this () type? where to use?

AsceticBear: Only in test now.

primitives/contract/src/mapping_token_factory/s2s.rs
```diff
@@ -67,7 +68,7 @@ impl Sub2SubMappingTokenFactory {
 /// @original_token: the origin token address
 /// @recipient: the final receiver of the token to be unlocked on the source chain
 /// @amount: the amount of the unlocked token
-#[derive(Debug, PartialEq, Eq)]
+#[derive(Encode, Decode, Debug, PartialEq, Eq, Clone)]
```
xiaoch05, line 71: why do we need to add this Encode and Decode feature?

AsceticBear: O, We don't need this after changing the recipient type to Vec. Great catch.

frame/wormhole/backing/s2s/src/lib.rs
AsceticBear: Done

frame/wormhole/issuing/s2s/src/lib.rs
AsceticBear: Done

frame/support/src/lib.rs
AsceticBear: Done

frame/wormhole/backing/s2s/src/lib.rs
```diff
@@ -360,56 +359,61 @@ pub mod pallet {
...
+				let message_id = nonce_to_message_id(&T::MessageLaneId::get(), message_nonce);
+				Self::deposit_event(Event::TokenUnlocked(
+					message_id,
+					token_address,
+					recipient_id,
+					amount,
+				));
+			} else {
+				return Err(<Error<T>>::InvalidRecipient.into());
+			}
```
xiaoch05, line 416: this can be write as ensure!(recipient.len() == 32, xxxx) not if else?

AsceticBear: Done

primitives/contract/src/mapping_token_factory/s2s.rs
AsceticBear: Done

xiaoch05 (approved):

primitives/s2s/src/call_params.rs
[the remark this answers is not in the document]
AsceticBear: It's not necessary indeed, include the `SourceChainAccountId: Encode`.

primitives/s2s/src/token_info.rs
```diff
@@ -0,0 +1,51 @@
...
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with Darwinia. If not, see <https://www.gnu.org/licenses/>.
+
+#![cfg_attr(not(feature = "std"), no_std)]
+
+pub const RING_NAME: &[u8] = b"Darwinia Network Native Token";
+pub const RING_SYMBOL: &[u8] = b"RING";
+pub const RING_DECIMAL: u8 = 9;
```
HackFisher, line 23: These should be config in runtime or delete later.

Will create a issue for this.

HackFisher (approved):

PR #910, opened 2021-11-05
Locked collateral for each order

AsceticBear:
Too many things are included in this pr, I made a list to help review.

Update List:
- Rename `AssignedRelayersStorage` to `AssignedRelayers` , **migration needed**.
- Rename `first_and_last_fee()` to `lowest_and_highest_fee`.
- Delete one config type `MiniumLockCollateral` and introduce another type `CollateralEachOrder`. The original `MiniumLockCollateral` was used to limit the minimum locked collateral when relayers enroll. At the same time, `MiniumLockCollateral` was the maximum slash amount for each assigned relayer when relaying delay occurred. A drawback this implementation has is how the market reward message relayer and confirm relayers for a delay message whose assigned relayers' original collateral had been slashed out. I made a change about this issue in this pull request. Introduce `order_capacity` for the relayer to indicate the ability to accept the order task. The capacity is calculated as `locked_collateral / CollateralEachOrder`. The relayer who has more collateral locked has more order_capacity, can eat more relaying orders. One thing that should be noted is the selection of assigned relayers also have a relationship with relayer order capacity.
- Slash logic update, the maximum slash value change to `CollateralEachOrder`. Once a message is delayed, slash `CollateralEachOrder` collateral. 

TODO:
- [x] Data Migration for `AssignedRelayers` and `Relayer`.
- [x] Benchmark test update
- [x] Update readme docs


AurevoirXavier:
Should we still keep `Fee<T>`?

BTW, Internal Review is not a good title.

frame/fee-market/src/lib.rs
```diff
@@ -215,20 +215,24 @@ pub mod pallet {
 				},
 				WithdrawReasons::all(),
 			);
-
-			<RelayersMap<T>>::insert(&who, Relayer::new(who.clone(), lock_collateral, fee));
-			<Relayers<T>>::append(who.clone());
+			// Store enrollment detail information.
+			<RelayersMap<T>>::insert(
+				&who,
+				Relayer::new(who.clone(), lock_collateral, fee, order_capacity),
```
HackFisher, line 221: `order_capacity` looks like a cache of `(lock_collateral / T::CollateralPerOrder::get()).saturated_into::<u32>()`, will it be better if we transform it to a pure internal function? The benefit is that the cached does not need be updated, and it will not be affected by the changes of `lock_collateral` or ` T::CollateralPerOrder::get()`.

frame/fee-market/src/lib.rs
```diff
@@ -241,53 +245,73 @@ pub mod pallet {
...
-		pub fn cancel_enrollment(origin: OriginFor<T>) -> DispatchResultWithPostInfo {
-			let who = ensure_signed(origin)?;
-			ensure!(Self::is_enrolled(&who), <Error<T>>::NotEnrolled);
+			let relayer = Self::relayer(&who);
 			ensure!(
-				<Relayers<T>>::get().len() > T::AssignedRelayersNumber::get() as usize,
-				<Error<T>>::TooFewEnrolledRelayers
+				new_collateral > relayer.collateral,
+				<Error<T>>::OnlyIncCollateralAllowed
 			);
```
HackFisher, line 253: In the new model, it should be ok to reduce collateral as soon as it does not affect the locked collateral in Orders.

CollateralForOrder * count_of_orders_in_delivering, or the sum of the collateral in delivering orders.

frame/fee-market/src/lib.rs
```diff
@@ -241,53 +245,73 @@ pub mod pallet {
...
-			ensure!(!Self::is_occupied(&who), <Error<T>>::OccupiedRelayer);
+			let old_capacity: u32 =
+				(relayer.collateral / T::CollateralPerOrder::get()).saturated_into::<u32>();
+			let used_capacity = old_capacity.saturating_sub(relayer.order_capacity);
+			let new_capacity: u32 =
+				(new_collateral / T::CollateralPerOrder::get()).saturated_into::<u32>();
```
HackFisher, line 258: cache updating, can be removed if we replace cache with calculation function.

frame/fee-market/src/lib.rs
AsceticBear: Yes, you can see it as a sign to the locked collateral. If we use a purely internal function, it may look like:
```rs
fn cal_order_capacity(who: & AccountId) -> u32 {
      Pallet::relayer(who).collateral / T::CollateralPerOrder::get()
}
```
Then, we have to call this function everywhere we. So I prefer the current implementation.

We only need to update this cache only in two cases, not so much in my view: 
1. Update locked collateral, a little complicated.
2. Slash collateral, slash CollateralPerOrder per message, order_capacity -1, this is easy to maintain.

frame/fee-market/src/lib.rs
AsceticBear: See the comments above.

frame/fee-market/src/lib.rs
AsceticBear: Right, we can remove the original `OnlyIncCollateralAllowed` limit.

frame/fee-market/src/lib.rs
HackFisher: > There are two hard things in computer science: cache invalidation, naming things, and off-by-one errors.

Don't underestimate the extra difficulty bring by cache management especially considering the audit burden and security requirement.

For the performance and ease of use, I can not tell difference between. Self.cal_order_capacity(account_id) and <RelayersMap<T>>::get(account_id).order_capacity

AsceticBear: Another thing is that `order_capacity` is useful to show the ability to relay when the relayer accepts an order or confirms an order. It's much harder to achieve this if we use the pure function. IMO.

frame/fee-market/src/lib.rs
AsceticBear: Now it supports decrease locked collateral. please review again.

frame/fee-market/src/lib.rs
```diff
@@ -68,14 +66,16 @@ pub mod pallet {
...
+
 		/// The minimum fee for relaying.
 		#[pallet::constant]
 		type MinimumRelayFee: Get<Fee<Self>>;
+		/// The assigned relayers number for each order.
 		#[pallet::constant]
 		type AssignedRelayersNumber: Get<u64>;
+		/// The collateral relayer need to lock for each order.
+		#[pallet::constant]
+		type CollateralPerOrder: Get<RingBalance<Self>>;
```
HackFisher, line 78: When we update this value in runtime upgrade, we might cause some invalidation in the correction of message capacity and collateral insufficiency. We might need to update the capacity in all relayers and invalidate some relayers(but some relayers might not be able to being removed due to working order)

One proposal will be just add the locked collateral value for each order in order struct, then there will not be issues for slashing when change collateralPerOrder. And capacity is calculated from collateral - all_locked_collateral_in_orders. (Whether or how to use cache is another question)

frame/fee-market/src/lib.rs
HackFisher: We can use rpc if the purpose is for ease of use for apps or bridger.

HackFisher (commented):
Willing to merge this earlier, but would like to resolve the conversations first, otherwise we will have to open another issue

frame/fee-market/src/lib.rs
AsceticBear: No, the apps or bridger can know the order_capacity thought RPC `get_relayer`. `The useful` means +1 when accepts an order in the callback, and -1 when confirms an order in the back. Then update_market select proper assgined relayers according to `order_capacity`.

See: 
1. https://github.com/darwinia-network/darwinia-common/blob/update-fee-market/frame/fee-market/src/s2s/callbacks.rs#L37
2. https://github.com/darwinia-network/darwinia-common/blob/update-fee-market/frame/fee-market/src/s2s/payment.rs#L173

AsceticBear:
You can take a review now @HackFisher . I am working on the remaining data migration works

AsceticBear:
Pangoro Try-runtime:

```sh
arwinia-common [update-fee-market] » ./target/release/drml try-runtime --chain pangoro-dev on-runtime-upgrade live wss://pangoro-rpc.darwinia.network
2021-11-11 16:54:10 `pallet_timestamp::UnixTime::now` is called at genesis, invalid value returned: 0    
2021-11-11 16:54:10 [#0] 🗳  Finalized election round with compute ElectionCompute::OnChain.    
2021-11-11 16:54:10 [0] 💸 new validator set of size 1 has been processed for era 1    
2021-11-11 16:54:10 initializing remote client to "wss://pangoro-rpc.darwinia.network"    
2021-11-11 16:54:12 scraping key-pairs from remote @ 0x2b58c83b03018e720bd576dcd0ec603798ea03a591f609dae80d605ce3bb94cf    
2021-11-11 16:54:12 downloading data for all modules.    
2021-11-11 16:54:37 Querying a total of 20124 keys    
2021-11-11 16:55:11 extending externalities with 1 manually injected keys    
2021-11-11 16:55:11 injecting a total of 20125 keys    
2021-11-11 16:55:11 Start migrate all storage items in AssignedRelayersStorage    
2021-11-11 16:55:11 the migrate content [Relayer { id: , collateral: 0, fee: 30000000000 }, Relayer { id: , collateral: 0, fee: 50000000000 }, Relayer { id: , collateral: 0, fee: 500000000000 }]    
2021-11-11 16:55:11 Migrate finished    
2021-11-11 16:55:11 ⚠️ System declares internal migrations (which *might* execute), setting storage version to <wasm:stripped>    
2021-11-11 16:55:11 ✅ no migration for FeeMarket, setting storage version to <wasm:stripped>    
2021-11-11 16:55:11 ✅ no migration for Substrate2SubstrateBacking, setting storage version to <wasm:stripped>    
2021-11-11 16:55:11 ✅ no migration for BridgePangolinGrandpa, setting storage version to <wasm:stripped>    
2021-11-11 16:55:11 ✅ no migration for BridgeDispatch, setting storage version to <wasm:stripped>    
2021-11-11 16:55:11 ✅ no migration for BridgePangolinMessages, setting storage version to <wasm:stripped>    
2021-11-11 16:55:11 ✅ no migration for Scheduler, setting storage version to <wasm:stripped>    
2021-11-11 16:55:11 ✅ no migration for Sudo, setting storage version to <wasm:stripped>    
2021-11-11 16:55:11 ✅ no migration for AuthorityDiscovery, setting storage version to <wasm:stripped>    
2021-11-11 16:55:11 ✅ no migration for ImOnline, setting storage version to <wasm:stripped>    
2021-11-11 16:55:11 ✅ no migration for Grandpa, setting storage version to <wasm:stripped>    
2021-11-11 16:55:11 ✅ no migration for Session, setting storage version to <wasm:stripped>    
2021-11-11 16:55:11 ✅ no migration for Historical, setting storage version to <wasm:stripped>    
2021-11-11 16:55:11 ⚠️ Offences declares internal migrations (which *might* execute), setting storage version to <wasm:stripped>    
2021-11-11 16:55:11 have 0 deferred offences, applying.    
2021-11-11 16:55:11 ✅ no migration for Staking, setting storage version to <wasm:stripped>    
2021-11-11 16:55:11 ✅ no migration for ElectionProviderMultiPhase, setting storage version to <wasm:stripped>    
2021-11-11 16:55:11 ✅ no migration for Authorship, setting storage version to <wasm:stripped>    
2021-11-11 16:55:11 ✅ no migration for TransactionPayment, setting storage version to <wasm:stripped>    
2021-11-11 16:55:11 ✅ no migration for Kton, setting storage version to <wasm:stripped>    
2021-11-11 16:55:11 ✅ no migration for Balances, setting storage version to <wasm:stripped>    
2021-11-11 16:55:11 ✅ no migration for Timestamp, setting storage version to <wasm:stripped>    
2021-11-11 16:55:11 ✅ no migration for Babe, setting storage version to <wasm:stripped>    
2021-11-11 16:55:11 ✅ no migration for RandomnessCollectiveFlip, setting storage version to <wasm:stripped>    
2021-11-11 16:55:11 TryRuntime_on_runtime_upgrade executed without errors. Consumed weight = 512620800000000, total weight = 512000000000000 (1.0012125) 
```

frame/fee-market/src/lib.rs
```diff
@@ -296,122 +312,109 @@ pub mod pallet {
...
 			WithdrawReasons::all(),
 		);
 		<RelayersMap<T>>::mutate(who.clone(), |relayer| {
 			relayer.collateral = new_collateral;
 		});
+
+		if Self::usable_order_capacity(&who) == 0 {
+			Self::remove_enrolled_relayer(who);
+			return;
+		}
```
HackFisher, line 358: Do not remove relayer here, let's relayer remove by themselves.

And this relay might have other order in working.

HackFisher (commented):
LGTM, except one more change request related to `remove_enrolled_relayer` in slashing.

frame/fee-market/src/lib.rs
AsceticBear: Done.

frame/fee-market/src/lib.rs
```diff
@@ -296,122 +312,104 @@ pub mod pallet {
...
+	/// Update relayer after slash occurred, this will changes RelayersMap storage. (Update market needed)
+	pub(crate) fn update_relayer_after_slash(who: &T::AccountId, new_collateral: RingBalance<T>) {
+		T::RingCurrency::set_lock(
 			T::LockId::get(),
-			who,
-			new_collateral,
+			&who,
+			LockFor::Common {
+				amount: new_collateral,
+			},
```
xiaoch05, line 348: maybe it's better we kick out the relayer directly. When the relayer's bridge does not work, the relayer should leave, otherwise, all the collateral will be lost.

frame/fee-market/src/s2s/payment.rs
```diff
@@ -188,61 +212,35 @@ where
...
+		locked_collateral >= slash_value,
+		"The locked collateral must alway greater than slash max"
+	);
+
+	let _ = <T as Config>::RingCurrency::transfer(
+		slash_account,
+		fund_account,
+		slash_value,
+		ExistenceRequirement::AllowDeath,
+	);
```
xiaoch05, line 233: should we compare the slash value and the left balance, and choose the minimize of them?

frame/fee-market/src/s2s/payment.rs
```diff
@@ -188,61 +212,35 @@ where
...
-		transfer_and_print_logs_on_error::<T>(slash_account, fund_account, slashed);
-		crate::Pallet::<T>::update_collateral(&slash_account, RingBalance::<T>::zero());
-	}
-	slashed
+	debug_assert!(
+		locked_collateral >= slash_value,
+		"The locked collateral must alway greater than slash max"
+	);
+
+	let _ = <T as Config>::RingCurrency::transfer(
```
xiaoch05, line 228: although we have to ignore this result. We should catch it and if failed, the returned `slash_value` should be Zero.

frame/fee-market/src/lib.rs
```diff
@@ -105,38 +103,34 @@ pub mod pallet {
...
 	pub enum Event<T: Config> {
 		/// Relayer enrollment
-		EnrollAndLockCollateral(T::AccountId, RingBalance<T>, Fee<T>),
-		/// Update relayer locked collateral
-		UpdateLockedCollateral(T::AccountId, RingBalance<T>),
-		/// Update relayer fee
-		UpdateRelayFee(T::AccountId, Fee<T>),
+		Enroll(T::AccountId, RingBalance<T>, Fee<T>),
+		/// Update relayer
+		UpdateRelayer(T::AccountId, Option<RingBalance<T>>, Option<Fee<T>>),
```
xiaoch05, line 108: suggest use two different Event to deposit Collateral update event and Fee update event

frame/fee-market/src/lib.rs
```diff
@@ -68,14 +66,16 @@ pub mod pallet {
 		type TreasuryPalletId: Get<PalletId>;
 		#[pallet::constant]
 		type LockId: Get<LockIdentifier>;
-		/// The minimum locked collateral for a fee market relayer, also represented as the maximum value for slash.
-		#[pallet::constant]
-		type MiniumLockCollateral: Get<RingBalance<Self>>;
+
 		/// The minimum fee for relaying.
 		#[pallet::constant]
 		type MinimumRelayFee: Get<Fee<Self>>;
```
xiaoch05, line 72: Type Fee<T> need to be removed

xiaoch05 (changes requested):

frame/fee-market/src/s2s/payment.rs
HackFisher: Return zero on Error looks strange too, prefer print error log if failed.

frame/fee-market/src/s2s/payment.rs
HackFisher: The design make sure that there are enough balance locked for slash, otherwise it it bug, so printing log might be enough for catch that bug.

frame/fee-market/src/s2s/payment.rs
xiaoch05: Later, this slash_value will be transferred from `fund_account` to the worked relayer. If we do not return Zero here, the `fund_account` will not have enough balance to give rewards.

frame/fee-market/src/lib.rs
HackFisher: > maybe it's better we kick out the relayer directly. When the relayer's bridge does not work, the relayer should leave, otherwise, all the collateral will be lost.

I don't agree that kicking out the relayer here as I mentioned in anther comments, there would be working orders that the relayer still need to handle, the slash is for some order, not for relayer.

And why all collateral will lost? Relayer can cancel his collateral later anytime.

frame/fee-market/src/s2s/payment.rs
```diff
@@ -139,47 +139,71 @@ where
...
+						// Total slash = message fee + assigned_relayers slash
+						// For each assigned relayer, slash CollateralPerOrder per order
+						let mut total_slash = message_fee;
+
+						let mut assigned_relayers_slash = RingBalance::<T>::zero();
+						for assigned_relayer in order.relayers_slice() {
+							let slashed = do_slash::<T>(
+								&assigned_relayer.id,
+								relayer_fund_account,
+								order.locked_collateral,
```
HackFisher, line 182: Oh, we are going to slash all the collateral in the order at once, not auction model?

I thought it is linear increasing till max value which equals order.locked_collateral.

frame/fee-market/src/s2s/payment.rs
HackFisher: > Later, this slash_value will be transferred from `fund_account` to the worked relayer. If we do not return Zero here, the `fund_account` will not have enough balance to pay rewards.

Yes, good catch.

And I also remember that there is slash and imbalance ways to avoid emit Transfer event here, should we avoid Transfer event here?

frame/fee-market/src/lib.rs
xiaoch05: yes this is a problem: the relayer can't be slashed by the other working orders again if kicked out.

all collateral will be lost because after being slashed. The left collateral will be locked again to accept next orders automatically.

HackFisher: It will not accept next orders automatically, accepting new order depends on capacity which is updated_collateral - all_orders_collateral.

frame/fee-market/src/s2s/payment.rs
xiaoch05: I think so, there is enough information to tell us where this slash balance goes. And this is in a callback function, it should be simple.

frame/fee-market/src/lib.rs
AsceticBear: > And why all collateral will lost? Relayer can cancel his enrollment and collateral later anytime.

It depends on whether this relayer has other in-process orders. If true,  he can't cancel the enrollment by himself. If the relaying bridge is dead, if this out-of-working time is long enough, all orders will be slashed after the relaying bridge works later.

```rs
ensure!(Self::occupied(&who).is_none(), <Error<T>>::OccupiedRelayer);
```

AsceticBear:
Latest Pangoro try-runtime:

```sh
darwinia-common [update-fee-market]× » ./target/release/drml try-runtime --block-at 0x02f8023395efc0a6ef163dfc2875ecd41018aeca14aae3cd9f958d31cee8c0f0 --chain pangoro-dev -u wss://pangoro-rpc.darwinia.network on-runtime-upgrade live
2021-11-12 11:55:45 `pallet_timestamp::UnixTime::now` is called at genesis, invalid value returned: 0    
2021-11-12 11:55:45 [#0] 🗳  Entering emergency mode: NoFallbackConfigured    
2021-11-12 11:55:45 [0] 💸 election provider failed due to NoFallbackConfigured    
2021-11-12 11:55:45 [#0] 🗳  Entering emergency mode: NoFallbackConfigured    
2021-11-12 11:55:45 [0] 💸 election provider failed due to NoFallbackConfigured    
2021-11-12 11:55:45 initializing remote client to "wss://pangoro-rpc.darwinia.network"    
2021-11-12 11:55:47 scraping key-pairs from remote @ 0x02f8023395efc0a6ef163dfc2875ecd41018aeca14aae3cd9f958d31cee8c0f0    
2021-11-12 11:55:47 downloading data for all modules.    
2021-11-12 11:56:12 Querying a total of 20156 keys    
2021-11-12 11:56:40 extending externalities with 1 manually injected key-values    
2021-11-12 11:56:40 injecting a total of 20158 keys    
2021-11-12 11:56:40 ===> Start migrate all storage items in AssignedRelayersStorage    
2021-11-12 11:56:40 the migrate content [Relayer { id: , collateral: 0, fee: 30000000000 }, Relayer { id: , collateral: 0, fee: 50000000000 }, Relayer { id: , collateral: 0, fee: 500000000000 }]    
2021-11-12 11:56:40 ===> End migrate all storage items in AssignedRelayersStorage    
2021-11-12 11:56:40 ===> Start migrate all storage items in Orders    
2021-11-12 11:56:40 The new order: order Order { lane: [114, 111, 108, 105], message: 62, sent_time: 1022299, confirm_time: None, locked_collateral: 100000000000, relayers: [PriorRelayer { id: , fee: 30000000000, valid_range: 1022299..1022349 }, PriorRelayer { id: , fee: 50000000000, valid_range: 1022349..1022399 }, PriorRelayer { id: , fee: 500000000000, valid_range: 1022399..1022449 }] }    
2021-11-12 11:56:40 ===> End migrate all storage items in Orders    
2021-11-12 11:56:40 ===> All Migrates finished    
2021-11-12 11:56:40 ⚠️ System declares internal migrations (which *might* execute), setting storage version to <wasm:stripped>    
2021-11-12 11:56:40 ✅ no migration for TransactionPause, setting storage version to <wasm:stripped>    
2021-11-12 11:56:40 ✅ no migration for FeeMarket, setting storage version to <wasm:stripped>    
2021-11-12 11:56:40 ✅ no migration for Substrate2SubstrateBacking, setting storage version to <wasm:stripped>    
2021-11-12 11:56:40 ✅ no migration for BridgePangolinGrandpa, setting storage version to <wasm:stripped>    
2021-11-12 11:56:40 ✅ no migration for BridgeDispatch, setting storage version to <wasm:stripped>    
2021-11-12 11:56:40 ✅ no migration for BridgePangolinMessages, setting storage version to <wasm:stripped>    
2021-11-12 11:56:40 ✅ no migration for Scheduler, setting storage version to <wasm:stripped>    
2021-11-12 11:56:40 ✅ no migration for Sudo, setting storage version to <wasm:stripped>    
2021-11-12 11:56:40 ✅ no migration for AuthorityDiscovery, setting storage version to <wasm:stripped>    
2021-11-12 11:56:40 ✅ no migration for ImOnline, setting storage version to <wasm:stripped>    
2021-11-12 11:56:40 ✅ no migration for Grandpa, setting storage version to <wasm:stripped>    
2021-11-12 11:56:40 ✅ no migration for Session, setting storage version to <wasm:stripped>    
2021-11-12 11:56:40 ✅ no migration for Historical, setting storage version to <wasm:stripped>    
2021-11-12 11:56:40 ⚠️ Offences declares internal migrations (which *might* execute), setting storage version to <wasm:stripped>    
2021-11-12 11:56:40 have 0 deferred offences, applying.    
2021-11-12 11:56:40 ✅ no migration for Staking, setting storage version to <wasm:stripped>    
2021-11-12 11:56:40 ✅ no migration for ElectionProviderMultiPhase, setting storage version to <wasm:stripped>    
2021-11-12 11:56:40 ✅ no migration for Authorship, setting storage version to <wasm:stripped>    
2021-11-12 11:56:40 ✅ no migration for TransactionPayment, setting storage version to <wasm:stripped>    
2021-11-12 11:56:40 ✅ no migration for Kton, setting storage version to <wasm:stripped>    
2021-11-12 11:56:40 ✅ no migration for Balances, setting storage version to <wasm:stripped>    
2021-11-12 11:56:40 ✅ no migration for Timestamp, setting storage version to <wasm:stripped>    
2021-11-12 11:56:40 ✅ no migration for Babe, setting storage version to <wasm:stripped>    
2021-11-12 11:56:40 ✅ no migration for RandomnessCollectiveFlip, setting storage version to <wasm:stripped>    
2021-11-12 11:56:40 TryRuntime_on_runtime_upgrade executed without errors. Consumed weight = 512646400000000, total weight = 512000000000000 (1.0012625)    

```

node/runtime/pangolin/src/lib.rs
```diff
@@ -800,23 +800,94 @@ impl dvm_rpc_runtime_api::ConvertTransaction<OpaqueExtrinsic> for TransactionCon
...
+			confirm_time: order.confirm_time,
+			locked_collateral: 100 * COIN,
+			relayers: order.relayers,
+		};
+		log::info!("The new order: order {:?}", new_order);
+		let hash = Blake2_128Concat::hash(&(new_order.lane, new_order.message).encode());
+		migration::put_storage_value(b"FeeMarket", b"Orders", &hash, new_order);
+	}
+	log::info!("===> End migrate all storage items in Orders");
+	log::info!("===> All Migrates finished");
```
AsceticBear, line 845: @AurevoirXavier Please help check the migration.

HackFisher:
Remove the `Fee` type in types.json

https://github.com/darwinia-network/darwinia-common/blob/7aa6d50126caece4fcd22fe01b4874d976ac8657/node/runtime/pangolin/types.json?_pjax=%23js-repo-pjax-container%2C%20div%5Bitemtype%3D%22http%3A%2F%2Fschema.org%2FSoftwareSourceCode%22%5D%20main%2C%20%5Bdata-pjax-container%5D#L261

and pangoro.

frame/fee-market/src/lib.rs
AsceticBear: Done

AsceticBear:
@HackFisher @xiaoch05 Take a review again.

frame/fee-market/src/lib.rs
```diff
@@ -151,22 +149,32 @@ pub mod pallet {
...
+		Order<T::AccountId, T::BlockNumber, RingBalance<T>>,
 		OptionQuery,
 	>;
 	#[pallet::storage]
 	pub type ConfirmedMessagesThisBlock<T: Config> =
 		StorageValue<_, Vec<(LaneId, MessageNonce)>, ValueQuery>;
 
+	#[pallet::storage]
+	#[pallet::getter(fn reward_mode)]
+	pub type RewardModeStorage<T: Config> =
```
HackFisher, line 171:
```suggestion
	pub type RewardMode<T: Config> =
```

AsceticBear: The enum is named `RewardMode`, So the storage can't the same as the enum.

HackFisher: Why not use `MaxSlashableCollateral` as I suggested, that would be more simple and clean.

AsceticBear: `MaxSlashableCollateral` is easy to confuse with an order's locked collateral(the max slash value for an order). Particularly we have introduced the S(t) function to calculate the conrete slash value.

frame/fee-market/src/s2s/payment.rs
```diff
@@ -212,8 +216,23 @@ where
...
+	if let Some(delay_blocks) = delay {
+		slash_value = UniqueSaturatedInto::<u128>::unique_saturated_into(delay_blocks)
+			.saturating_mul(UniqueSaturatedInto::<u128>::unique_saturated_into(
+				T::SlashPerBlock::get(),
+			))
+			.unique_saturated_into();
+	}
+	sp_std::cmp::min(locked_collateral, slash_value)
+}
+
```
HackFisher: This should be a trait that can be configured in runtime, we might want to testing new `S(t)` in tests or testnet.

AsceticBear: I see, just like the old slasher associated type. Let me give it a change.

frame/fee-market/src/lib.rs
HackFisher: It is naming issue, let's choose a better name then.

How about `CollateralSlashProtect`

frame/fee-market/src/s2s/payment.rs
HackFisher: > Later, this slash_value will be transferred from `fund_account` to the worked relayer. If we do not return Zero here, the `fund_account` will not have enough balance to pay rewards.

@AsceticBear The issue @xiaoch05 mentioned here is an important security issue, not resolved yet. The return value should not be the slash_value but zero if transfer failed.

frame/fee-market/src/lib.rs
AsceticBear: I prefer `CollateralSlashMode`, In normal cases, nothing related to protection, right?

frame/fee-market/src/s2s/payment.rs
AsceticBear: Done

frame/fee-market/src/s2s/payment.rs
AsceticBear: Done

frame/fee-market/src/lib.rs
AsceticBear: Done

frame/fee-market/src/lib.rs
```diff
@@ -101,42 +100,41 @@ pub mod pallet {
...
+		Enroll(T::AccountId, RingBalance<T>, RingBalance<T>),
 		/// Update relayer locked collateral
 		UpdateLockedCollateral(T::AccountId, RingBalance<T>),
 		/// Update relayer fee
-		UpdateRelayFee(T::AccountId, Fee<T>),
+		UpdateRelayFee(T::AccountId, RingBalance<T>),
 		/// Relayer cancel enrollment
 		CancelEnrollment(T::AccountId),
+		/// Update collateral slash protect value
+		UpdateCollateralSlashProtect(RingBalance<T>),
```
xiaoch05, line 114: need more docs about the params of the above events.

frame/fee-market/src/lib.rs
```diff
@@ -296,79 +328,70 @@ pub mod pallet {
...
+	pub(crate) fn remove_enrolled_relayer(who: &T::AccountId) {
 		T::RingCurrency::remove_lock(T::LockId::get(), who);
+
 		<RelayersMap<T>>::remove(who.clone());
 		<Relayers<T>>::mutate(|relayers| relayers.retain(|x| x != who));
+		<AssignedRelayers<T>>::mutate(|assigned_relayers| {
+			if let Some(relayers) = assigned_relayers {
+				relayers.retain(|x| x.id != *who);
+			}
+		});
```
HackFisher, line 384: In which cases we will remove an relayer who is in assignment? Should we change this to use `ensure!`?

AsceticBear: We need to update the assigned_relayers list when the enrolled relayer wants to exit the fee market by calling` cancel_enrollment()`.

HackFisher: But  `assigned_relayers` means there exists order for the relayer right, we do not allow cancel enrollment for the relayer if it has ongoing order?

AsceticBear: > But assigned_relayers means there exists order for the relayer right?

we use `occupied(who: &T::AccountId)`.

> we do not allow cancel enrollment for the relayer if it has ongoing order?

Yes,  we will check whether the relayer have ongoing orders use `ensure!(Self::occupied(&who).is_none(), <Error<T>>::OccupiedRelayer);` before `remove_enrolled_relayer`.

HackFisher (approved):

AurevoirXavier:
@sxlwar Types updated.

PR #919, opened 2021-11-10
Optimization for precompile contract

xiaoch05:
- use a common macro `darwinia_evm_precompile_utils::selector` to define the selector from contract
- split misc to s2s and ethereum bridge precompile contracts
- add abi encode for the result of the precompile contract and abi decode for the params


frame/evm/precompile/contracts/utils/Cargo.toml
```diff
@@ -0,0 +1,23 @@
...
+homepage    = "https://darwinia.network/"
+license     = "GPL-3.0"
+name        = "darwinia-evm-precompile-utils"
+readme      = "README.md"
+repository  = "https://github.com/darwinia-network/darwinia-common/"
+version     = "2.6.9"
+
+[dependencies]
+# crates.io
+evm   = { version = "0.27.0", default-features = false, features = ["with-codec"] }
```
AurevoirXavier, line 14:
```suggest
evm   = { version = "0.27", default-features = false, features = ["with-codec"] }
```

AurevoirXavier:
I think we don't need a proc macro.

xiaoch05:
> I think we don't need a proc macro.

what's your suggestion about it?

HackFisher (approved):

AurevoirXavier:
> what's your suggestion about it?

Let me try.

frame/evm/precompile/contracts/bridge/ethereum/src/lib.rs
```diff
@@ -0,0 +1,79 @@
...
+use darwinia_evm_precompile_utils::{selector, DvmInputParser};
+use dp_evm::Precompile;
+
+#[selector]
+enum Action {
+	BurnAndRemoteUnlock = "burn_and_remote_unlock(uint32,address,address,address,bytes,uint256)",
+	TokenRegisterResponse = "token_register_response(address,address,address)",
+}
+
+/// The contract address: 0000000000000000000000000000000000000017
```
HackFisher, line 36: Not in the scope of this PR, but we still need to think about the management of precompile address allocation and compatibility design.

These actions looks short life, there could be more similar actions from business level in future, and also this could be removed too in future.

AsceticBear:
Do not merge before I take a review, thanks.

AurevoirXavier:
I'm not sure if this magic number should be written explicitly.
@xiaoch05 said it's okay.

```rust
enum Action {
    BurnAndRemoteUnlock = 744680220,
    TokenRegisterResponse = 2054786895,
}
impl Action {
    pub fn from_u32(value: u32) -> Result<Self, ExitError> {
        match value {
            744680220 => Ok(Action::BurnAndRemoteUnlock),
            2054786895 => Ok(Action::TokenRegisterResponse),
            _ => Err(ExitError::Other("mismatch the enum value".into())),
        }
    }
}
```

HackFisher:
> I'm not sure if this magic number should be written explicitly. @xiaoch05 said it's okay.
> 
> ```rust
> enum Action {
>     BurnAndRemoteUnlock = 744680220,
>     TokenRegisterResponse = 2054786895,
> }
> impl Action {
>     pub fn from_u32(value: u32) -> Result<Self, ExitError> {
>         match value {
>             744680220 => Ok(Action::BurnAndRemoteUnlock),
>             2054786895 => Ok(Action::TokenRegisterResponse),
>             _ => Err(ExitError::Other("mismatch the enum value".into())),
>         }
>     }
> }
> ```

How this numbers generated? Are they random?

xiaoch05:
> > I'm not sure if this magic number should be written explicitly. @xiaoch05 said it's okay.
> > ```rust
> > enum Action {
> >     BurnAndRemoteUnlock = 744680220,
> >     TokenRegisterResponse = 2054786895,
> > }
> > impl Action {
> >     pub fn from_u32(value: u32) -> Result<Self, ExitError> {
> >         match value {
> >             744680220 => Ok(Action::BurnAndRemoteUnlock),
> >             2054786895 => Ok(Action::TokenRegisterResponse),
> >             _ => Err(ExitError::Other("mismatch the enum value".into())),
> >         }
> >     }
> > }
> > ```
> 
> How these numbers generated? Are they random?

It's not random, it's the func's signature. 4 bytes = u32

AurevoirXavier (approved):

frame/evm/precompile/contracts/utils/src/lib.rs
```diff
@@ -0,0 +1,28 @@
+#![cfg_attr(not(feature = "std"), no_std)]
+
+pub use darwinia_evm_precompile_utils_macro::selector;
+use evm::ExitError;
+
+const ACTION_LEN: usize = 4;
```
AsceticBear, line 6: Consider using https://github.com/darwinia-network/darwinia-common/blob/master/frame/support/src/evm.rs#L33. The same as `ACTION_LEN` here.

frame/evm/precompile/contracts/utils/src/lib.rs
```diff
@@ -0,0 +1,28 @@
+#![cfg_attr(not(feature = "std"), no_std)]
```
AsceticBear, line 1: License missing for this crate.

frame/evm/precompile/contracts/utils/src/lib.rs
xiaoch05: move this here? This is only used in precompile contract.

AsceticBear: Just reuse it. It was used in the `transfer` precompile before.

xiaoch05: transfer precompile is also precompile contract? so it's better to put it in precompile/contracts/utils?

AsceticBear: the `transfer` is different from the two precompiles here. Just leave it there untouch.

xiaoch05: ok

AsceticBear:
LGTM. Only one comment, we just keep one selector type is enough.

AsceticBear (approved):

PR #922, opened 2021-11-11
Add RPC to query orders in progress

### LINKED ISSUE #912: Support query all created orders in the fee market
Currently, we can only query created order detailed info by (lane_id, message_nonce)in Apps. if you don't know(or not sure) the exact message nonce or lane_id, no way to check out the order.

One proposal is adding RPC interfaces like `market_fee()`, named `orders()`. This interface is useful in case of debugging order-related issues.

AsceticBear:
Close #912


AurevoirXavier (changes requested):
Please add the new types in types.json's RPC part.

AurevoirXavier (approved):

HackFisher (approved):

PR #927, opened 2021-11-15
Release v2.6.10

AurevoirXavier:


node/runtime/pangoro/src/lib.rs
```diff
@@ -135,7 +135,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
 	spec_name: sp_runtime::create_runtime_str!("Pangoro"),
 	impl_name: sp_runtime::create_runtime_str!("Pangoro"),
 	authoring_version: 1,
-	spec_version: 2690,
+	spec_version: 26100,
```
HackFisher, line 138: This breaks my understanding of the spec version and crate version.

How about using 20610 instead, compared to 2.6.10, aka. 02.06.10

node/runtime/pangolin/src/lib.rs
```diff
@@ -171,7 +171,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
 	spec_name: sp_runtime::create_runtime_str!("Pangolin"),
 	impl_name: sp_runtime::create_runtime_str!("Pangolin"),
 	authoring_version: 1,
-	spec_version: 2690,
+	spec_version: 26100,
```
HackFisher, line 174:
```suggestion
	spec_version: 20610,
```

xiaoch05 (approved):

node/runtime/pangoro/src/lib.rs
HackFisher: What will the spec version be if the crate is 2.11.9?

AurevoirXavier: Extend to `2 11 900`

HackFisher: 9.10.5   91050
10.9.5  10950

Here is an example which doesn't make sense to me, 9.10.5 is smaller than 10.9.5, but 91050 is larger than 10950

AurevoirXavier: Once we extend any of these fields we won't shrink it in the future.

AurevoirXavier: And I'll check these numbers in CI.

HackFisher: I know, but I'm also talking about human readability of spec version, e.g. when we encounter spec version 11240, which crate number we expect? 11.2.4 or 1.12.4?

AurevoirXavier: I just format it check out the latest commit.

HackFisher: > I have just format it check out the latest commit ;)

This only resolve the format of the rust code, not other places such as apps and subscan.

But some recap & revise to the convention will help me to accept current change:

We allocate digits length to each version, including MAJOR.MINOR.PATCH-Prerelease, now allocate (1,1,2,1) for them, once we increase the digits length for one of them, should not be reduced later. For example, if we have a version 2.7.0 next, we should use 27000 instead of 2700, digits length is (1,1,2,1) not back to (1,1,1,1)

AurevoirXavier: Yep. The length should be kept.

---

And I think once outside rust code, people won't care about the crate version anymore.
So, in Apps users just need to know this version has been upgraded.

HackFisher (approved):

PR #935, opened 2021-11-16
Update `AssignedRelayersNumber` in pangoro

AsceticBear:
Keep the same as Pangolin. Reducing the burden of testnet maintenance.


fewensa (approved):

AurevoirXavier:
Why not make this a storage?

AsceticBear:
> Why not make this a storage?

Make sense. We can update this value without a runtime upgrade in this way.

frame/fee-market/src/lib.rs
```diff
@@ -170,6 +169,15 @@ pub mod pallet {
 	#[pallet::getter(fn collateral_slash_protect)]
 	pub type CollateralSlashProtect<T: Config> = StorageValue<_, RingBalance<T>, OptionQuery>;
 
+	#[pallet::storage]
+	#[pallet::getter(fn assigned_relayers_number)]
+	pub type AssignedRelayersNumber<T: Config> =
+		StorageValue<_, u32, ValueQuery, DefaultAssignedRelayersNumber>;
+	#[pallet::type_value]
+	pub fn DefaultAssignedRelayersNumber() -> u32 {
```
AurevoirXavier, line 177: Interesting.

AurevoirXavier (approved):

PR #939, opened 2021-11-16
Reduce the callbacks weight used

### LINKED ISSUE #911: More accurate weight consume needed.
Now the `FeeMarketMessageConfirmedHandler` return hardcoded weight `reads_writes(1, 1)` instead of the consumed weight, because of the callback limitations for each relay message. 

https://github.com/paritytech/parity-bridges-common/blob/master/modules/messages/src/lib.rs#L612

```rs
pub struct FeeMarketMessageConfirmedHandler<T>(PhantomData<T>);

impl<T: Config> OnDeliveryConfirmed for FeeMarketMessageConfirmedHandler<T> {
	fn on_messages_delivered(lane: &LaneId, delivered_messages: &DeliveredMessages) -> Weight {
		let now = frame_system::Pallet::<T>::block_number();
		for message_nonce in delivered_messages.begin..=delivered_messages.end {
			if let Some(order) = <Orders<T>>::get((lane, message_nonce)) {
				if !order.is_confirmed() {
					<Orders<T>>::mutate((lane, message_nonce), |order| match order {
						Some(order) => order.set_confirm_time(Some(now)),
						None => {}
					});
					<ConfirmedMessagesThisBlock<T>>::append((lane, message_nonce));
				}
			}
		}

		<T as frame_system::Config>::DbWeight::get().reads_writes(1, 1)
	}
}

```

AsceticBear:
Close #911. By removing `ConfirmedMessagesThisBlock` storage to fit the callback weight limit.


frame/fee-market/src/s2s/payment.rs
```diff
@@ -123,88 +130,89 @@ where
...
-			match order.lowest_and_highest_fee() {
-				(Some(lowest_fee), Some(message_fee)) => {
-					let message_reward;
-					let confirm_reward;
+		for message_nonce in nonce_begin..nonce_end + 1 {
+			// The order created when message was accepted, so we can always get the order info below.
+			if let Some(order) = <Orders<T>>::get(&(lane_id, message_nonce)) {
+				// The confirm_time of the order is set in the `OnDeliveryConfirmed` callback. And the callback function
+				// was called as source chain received message delivery proof, before the reward payment.
+				let order_confirm_time = order
```
HackFisher, line 142: @AsceticBear Here, order's confirm_time is supposed to be modified with order_confirm_time right? But I can not find the modification code.

frame/fee-market/src/lib.rs
```diff
@@ -184,10 +181,11 @@ pub mod pallet {
 	impl<T: Config> Hooks<BlockNumberFor<T>> for Pallet<T> {
 		fn on_finalize(_: BlockNumberFor<T>) {
 			// Clean the order's storage when the rewards has been paid off
-			for (lane_id, message_nonce) in <ConfirmedMessagesThisBlock<T>>::get() {
-				<Orders<T>>::remove((lane_id, message_nonce));
+			for ((lane_id, message_nonce), order) in <Orders<T>>::iter() {
+				if order.confirm_time.is_some() {
```
HackFisher, line 185: Add documents, if confirm_time is some, meaning this order has already been processed.

frame/fee-market/src/s2s/payment.rs
AsceticBear: We modified the confirmed time in the callback. See https://github.com/darwinia-network/darwinia-common/blob/change-code-to-adapt-upstream/frame/fee-market/src/s2s/callbacks.rs#L60

HackFisher (approved):

PR #947, opened 2021-11-20
Optimize CI

AurevoirXavier:
1. Optimize Caching
1. Save the cost of GCS
1. Speed the compiling time up to 20 mins


AurevoirXavier:
[BOT CHECK] Code, Runtime

AurevoirXavier:
1. Save the cost of GCS
2. Speed up the compiling time to 20 mins

PR #948, opened 2021-11-21
Correct Dispatch Instance

AurevoirXavier:


node/runtime/pangolin/src/lib.rs
```diff
@@ -280,7 +280,7 @@ frame_support::construct_runtime! {
 		// DynamicFee: dvm_dynamic_fee::{Pallet, Call, Storage, Inherent} = 47,
 
 		BridgePangoroMessages: pallet_bridge_messages::<Instance1>::{Pallet, Call, Storage, Event<T>} = 43,
-		BridgeDispatch: pallet_bridge_dispatch::{Pallet, Event<T>} = 44,
+		BridgeDispatch: pallet_bridge_dispatch::<Instance1>::{Pallet, Event<T>} = 44,
```
HackFisher, line 283: Why we need to change to use `Instance1`, I thought BridgeDispatch can be used by multiple bridge instances, we only need on bridge dispatch pallet.

AurevoirXavier: Yes, I made a mistake here. 
That's why I mark this as draft PR yesterday.
Because the Instances name in here is a mess.

I'll update them in #949.

PR #951, opened 2021-11-22
Bump deps

fewensa:
If there is no problem with https://github.com/darwinia-network/parity-bridges-common/pull/8, I will create a new tag and modify the dependencies here 

@HackFisher @AurevoirXavier


AurevoirXavier:
Why not use a tag?

fewensa:
Please review https://github.com/darwinia-network/parity-bridges-common/pull/8 first, If there is no problem, I will create a new tag

HackFisher (approved):

PR #956, opened 2021-11-24
Improve CI

AurevoirXavier:
1. Only running on code/genesis/ci changed.
2. Use matrix.
3. Optimize cache.
4. Remove compress step.


HackFisher (approved):
LGTM, but required checks in github settings need to be updated.

PR #958, opened 2021-11-26
Improve DVM Tests

AsceticBear:


AsceticBear:
I need to mark this pull request status to `Ready for review` to test ci running. I'll add comments in this pr when it's truly ready for review.

AurevoirXavier:
Why we need a new node template?

AsceticBear:
> Why do we need a new node template?

Because these tests are based on the node `manual-seal` blocking mechanism, which pangolin does not need. I have tried to add `manual-seal` to pangolin via feature control, but this approach would increase the complexity of maintaining pangolin. So I prefer to add a separate, very simple node template, which has the following benefits:
1. only provide dvm pallets support for testing, which is simple enough. 
2. no relation to pangolin
3. good for keeping in sync with upstream

AurevoirXavier:
`manual-seal` is based on AURA.
Does this test suitable for BABE?

AsceticBear:
> `manual-seal` is based on AURA. Does this test suitable for BABE?

Not well, Some tests rely on the `manual-seal` feature.

AurevoirXavier:
But Darwinia/Crab are using BABE as the consensus engine. Are these tests useful for us?

AsceticBear:
Absolutely useful. Most of the tests are designed to ensure the correct behavior of DVM pallets and clients. 

There seems to be some contradiction in what is happening here. The test point has nothing to do with the consensus block out mechanism itself, but the test design wants to be able to control the speed of the block out. For example, send a transaction, but expect it to pack into the next block, expect to verify the contents of the next block, etc. In the case of babe, it is possible that the transaction will be packed after n blocks and will keep coming out of the block.

tests/ts-tests/template/node/src/service.rs
```diff
@@ -0,0 +1,740 @@
...
+							},
+						});
+					// we spawn the future on a background thread managed by service.
+					task_manager
+						.spawn_essential_handle()
+						.spawn_blocking("instant-seal", authorship_future);
+				}
+			};
+		}
+		log::info!("Manual Seal Ready");
```
AurevoirXavier, line 485: Looks like whatever the sealing is. It will always print `Manual Seal Ready`.

AurevoirXavier:
Once we make some substrate updates or frontier's.
It's hard to make a companion for this external node.
And the CI needs to build 2 nodes.
So I add a template node to drml as a preset of this update.

tests/dvm/.gitignore
```diff
@@ -1,3 +1,3 @@
 node_modules/
 build/
-target/
+package.json
```
AurevoirXavier, line 3: I think this file can be removed.

There is .gitignore in the root which will handle these.

AsceticBear: Okay.

AurevoirXavier:
Frontier test in local:
```
TSError: ⨯ Unable to compile TypeScript:
tests/util.ts:2:33 - error TS2307: Cannot find module 'web3-core-helpers' or its corresponding type declarations.

2 import { JsonRpcResponse } from "web3-core-helpers";
```

Btw, is necessary to build the solc from docker? cc @hujw77

hujw77:
> Frontier test in local:
> 
> ```
> TSError: ⨯ Unable to compile TypeScript:
> tests/util.ts:2:33 - error TS2307: Cannot find module 'web3-core-helpers' or its corresponding type declarations.
> 
> 2 import { JsonRpcResponse } from "web3-core-helpers";
> ```
> 
> Btw, is necessary to build the solc from docker? cc @hujw77

Not necessary, if you use js/ts, `solcjs` may be helpful.
And I am not like docker personally.

AurevoirXavier:
> Not necessary, if you use js/ts, `solcjs` may be helpful. And I am not like docker personally.

Could you help me to update it?

hujw77:
> > Not necessary, if you use js/ts, `solcjs` may be helpful. And I am not like docker personally.
> 
> Could you help me to update this?

OK

PR #965, opened 2021-12-01
Update reward order base fee

### LINKED ISSUE #962: Improve fee-market reward distribution
```
order sent_time: 100, slot = 50 => Order relayer 1: [100-150] relayer2: [150-200] relayer3: [200-250]

Reward part(payment.rs): confirm_time: [180] -> r2

60% p1 -> assigned relayer r2
80% * (40% * p1) -> message relayer
20% * (40% * p1) -> confirm relayer
p3 - p1 -> treasury
```
For now, If message `confirm_time` is in T1 ~ T3, reward base amount is always `p1`.
I think it is more reasonable to use the `relayer` required price `Pn` as the reward base amount.
In the case of the above example, p2 instead of p1 is more appreciate.

AsceticBear:
Close #962


hujw77 (commented):
LGTM

HackFisher (approved):

PR #986, opened 2021-12-07
Fix failed ts tests

hujw77:
- Fix Path
- Use Release
- With `sccache`
- Fix ENV
- Fix ENV
- Fix Permission
- Fix Permission
- Allow Create Empty Block
- Add `create_empty` param at `createAndFinalizeBlock`
- Just try to set `create_empty` to false at `createAndFinalizeBlock`
- Use `ethers` instead of `web3`
- Empty
- rm empty


HackFisher:
Why ci is not running?

fewensa:
https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#onpushpull_requestpaths

https://github.com/darwinia-network/darwinia-common/blob/2e267ef6b50218528e07fd42b8d296a5a995566f/.github/workflows/ci.yml#L18-L23


Not support running when modified ts file

AsceticBear:
@hujw77 @AurevoirXavier Please take a review later. I think it is much more stable after running several times without error.

AurevoirXavier (approved):

hujw77 (commented):
LGTM

HackFisher (approved):

PR #1024, opened 2022-01-05
Code Cleaning and Fix Relayer Ord

AurevoirXavier:


primitives/fee-market/src/lib.rs
```diff
@@ -191,8 +187,7 @@ pub struct PriorRelayer<AccountId, BlockNumber, Balance> {
 	pub fee: Balance,
 	pub valid_range: Range<BlockNumber>,
 }
-
-impl<'a, AccountId, BlockNumber, Balance> PriorRelayer<AccountId, BlockNumber, Balance>
+impl<AccountId, BlockNumber, Balance> PriorRelayer<AccountId, BlockNumber, Balance>
 where
 	BlockNumber: sp_std::ops::Add<Output = BlockNumber> + Clone,
```
AsceticBear, line 192: Please help change this line to `BlockNumber: AtLeast32BitUnsigned + Copy` by the way.

AsceticBear (changes requested):

primitives/fee-market/src/lib.rs
```diff
@@ -42,43 +41,39 @@ impl<AccountId, Balance> Relayer<AccountId, Balance> {
...
 	fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
-		if self.fee == other.fee {
-			return other.collateral.partial_cmp(&self.collateral);
-		}
-		self.fee.partial_cmp(&other.fee)
+		Some(match self.fee.cmp(&other.fee) {
+			// We reverse the order here to turn the collateral value into rank.
+			//
+			// Use `other.cmp(self)` instead of `self.cmp(other)`.
+			Ordering::Equal => other.collateral.cmp(&self.collateral),
```
AsceticBear, line 54: Why not `partial_cmp` here?  Actually, We do not need a match here, `==` equal to `Ordering::Equal`, just like the original style.

AurevoirXavier: 1. Use `cmp` is more logical
2. Less calculating works, the origin code call `cmp` twice

AsceticBear (approved):

PR #1026, opened 2022-01-06
Copyright 2021 -> 2022

AsceticBear:


AurevoirXavier:
Should we modify PureStake's copyright?

AsceticBear:
> Should we modify PureStake's copyright?

I think it would be ok. They are upgrading copyright now. https://github.com/PureStake/moonbeam/pull/1146

PR #1027, opened 2022-01-07
Initialize account nonce in building genesis

AsceticBear:
When initializing some accounts in the Genesis build, we should increase account nonce as well as balance.


AurevoirXavier (approved):

AurevoirXavier:
Just thought about that do we need a migration?

PR #1029, opened 2022-01-08
Replace DVM modules with frontier(Part I)

AsceticBear:
without any logical changes.

This is the first part of this replacement task, We can replace more after the darwinia-frontier branch's `ethereum` dep update to 0.10.0, which supports metadata v14. I would like to raise separated pull requests to make review easier.


AurevoirXavier (approved):
GJ

PR #1031, opened 2022-01-10
DVM signed transaction

AsceticBear:
Companion for https://github.com/paritytech/frontier/pull/482 and https://github.com/paritytech/frontier/pull/495.

We rely on the darwinia-frontier branch 0.12.1(The main branch is darwinia-frontier 0.12.0). Might merge these changes to 0.12.0 and upgrade to pangolin directly.

### DIFF
--- a/frame/dvm/src/lib.rs
+++ b/frame/dvm/src/lib.rs
@@ -52,23 +52,26 @@ use frame_support::storage::unhashed;
 use frame_support::{
 	dispatch::DispatchResultWithPostInfo,
 	ensure,
-	traits::{Currency, Get},
+	traits::{Currency, EnsureOrigin, Get},
 	weights::{Pays, PostDispatchInfo, Weight},
 	PalletId,
 };
-use frame_system::ensure_none;
+use frame_system::pallet_prelude::OriginFor;
+use scale_info::TypeInfo;
 use sp_runtime::{
 	generic::DigestItem,
 	traits::{One, Saturating, UniqueSaturatedInto, Zero},
 	transaction_validity::{
-		InvalidTransaction, TransactionSource, TransactionValidity, ValidTransactionBuilder,
+		InvalidTransaction, TransactionValidity, TransactionValidityError, ValidTransactionBuilder,
 	},
-	DispatchError,
+	DispatchError, RuntimeDebug,
 };
 use sp_std::{marker::PhantomData, prelude::*};
 // --- darwinia-network ---
 use darwinia_evm::{AccountBasic, BlockHashMapping, FeeCalculator, GasWeightMapping, Runner};
-use darwinia_support::evm::{recover_signer, DVMTransaction, IntoH160, INTERNAL_TX_GAS_LIMIT};
+use darwinia_support::evm::{
+	new_internal_transaction, recover_signer, DVMTransaction, IntoH160, INTERNAL_TX_GAS_LIMIT,
+};
 use dp_consensus::{PostLog, PreLog, FRONTIER_ENGINE_ID};
 
 /// A type alias for the balance type from this pallet's point of view.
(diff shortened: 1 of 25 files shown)


frame/dvm/src/lib.rs
```diff
@@ -332,37 +349,154 @@ pub mod pallet {
...
+			.and_provides((origin, transaction.nonce))
+			.priority(transaction.gas_price.unique_saturated_into());
+
+		// In the context of the pool, a transaction with
+		// too high a nonce is still considered valid
+		if transaction.nonce > account_nonce {
+			if let Some(prev_nonce) = transaction.nonce.checked_sub(1.into()) {
+				builder = builder.and_requires((origin, prev_nonce))
+			}
+		}
```
xiaoch05, line 432: do we need to limit this nonce?
If users send so many txs that nonce very large. Whether the transaction pool will drop it?

frame/dvm/src/lib.rs
```diff
@@ -332,37 +349,154 @@ pub mod pallet {
...
 			.into())
 		})?
 	}
 
-	/// Execute DVMTransaction in evm runner and save the execution info in Pending
-	fn raw_transact(transaction: DVMTransaction) -> Result<(ExitReason, U256), DispatchError> {
+	// Execute Transaction in evm runner and save the execution info in Pending
+	fn raw_transact(
+		source: H160,
+		dvm_transaction: DVMTransaction,
```
xiaoch05, line 481: After we separate the source from DVMTransaction,  do we need this struct anymore? or use TransactionV0 directly?

xiaoch05, line 481: After we separate the source from DVMTransaction,  do we need this struct anymore? or use TransactionV0 directly?

frame/dvm/src/lib.rs
```diff
@@ -78,6 +81,91 @@ type KtonCurrency<T> = <T as Config>::KtonCurrency;
...
+	EthereumTransaction(H160),
+}
+
+pub fn ensure_ethereum_transaction<OuterOrigin>(o: OuterOrigin) -> Result<H160, &'static str>
+where
+	OuterOrigin: Into<Result<RawOrigin, OuterOrigin>>,
+{
+	match o.into() {
+		Ok(RawOrigin::EthereumTransaction(n)) => Ok(n),
+		_ => Err("bad origin: expected to be an Ethereum transaction"),
```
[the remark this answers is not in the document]
AsceticBear: I'd prefer to keep this consistent with the upstream repo now. Or we can create a pull request in frontier firstly.

frame/dvm/src/lib.rs
AsceticBear: No, we don't need to do this. Here we use substrate tx pool, the limit task should be controlled there.

frame/dvm/src/lib.rs
AsceticBear: We can't do this, because `gas_price` should be none in the internal transact execute process.  And we can't even set the gas_price=0, this brings execution error in the evm runner.

AsceticBear: The same with above. because of `gas_price`

frame/dvm/src/lib.rs
xiaoch05: If I attack the tx pool by sending so many txs with invalid nonce(nonce > currentNonce) continuously. How can u process the pool？

AsceticBear: 1. If all the txs with the same nonce(large than the current one), will trigger the `TemporarilyBanned` mechanism, then you have to stop your attack behavior.
2. If all the txs with different nonce(increase one by one), there is a max limit for transaction future pool, If reached, then trigger enforce clean up. 
- https://github.com/paritytech/substrate/blob/master/client/transaction-pool/src/graph/validated_pool.rs#L239
- https://github.com/paritytech/substrate/blob/master/client/transaction-pool/src/graph/pool.rs#L113

xiaoch05 (approved):

HackFisher (approved):

hujw77 (commented):
LGTM

PR #1037, opened 2022-01-12
Build for nixos support

hujw77:
- Add shell.nix
- Add build for nixos support
- Fix typo

### DIFF
--- a/README.md
+++ b/README.md
@@ -99,5 +99,5 @@ target/release/drml \
 #### Build in Nixos
 ```[shell]
 nix-shell
-cargo build [-p drml] [--relese]
+cargo build [-p drml] [--release]
 ```
--- a/README.md
+++ b/README.md
@@ -100,5 +100,5 @@ target/release/drml \
 ### NixOS
 ```sh
 nix-shell
-cargo build [-p drml] [--release]
+cargo build [-p drml] [--release] [--features template]
 ```
--- a/README.md
+++ b/README.md
@@ -95,3 +95,9 @@ target/release/drml \
 	--alice \
 	--base-path tests/local-testnet/alice
 ```
+
+#### Build in Nixos
+```[shell]
+nix-shell
+cargo build [-p drml] [--relese]
+```
--- a/README.md
+++ b/README.md
@@ -98,7 +98,7 @@ target/release/drml \
 
 ## Build
 ### NixOS
-```[shell]
+```sh
 nix-shell
 cargo build [-p drml] [--release]
 ```
--- a/README.md
+++ b/README.md
@@ -97,7 +97,7 @@ target/release/drml \
 ```
 
 ## Build
-### NixOs
+### NixOS
 ```[shell]
 nix-shell
 cargo build [-p drml] [--release]
--- a/README.md
+++ b/README.md
@@ -97,7 +97,7 @@ target/release/drml \
 ```
 
 ## Build
-### Nixos
+### NixOs
 ```[shell]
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -95,3 +95,9 @@ target/release/drml \
 	--alice \
 	--base-path tests/local-testnet/alice
 ```
+
+#### Build in Nixos
```
AurevoirXavier, line 99: This part is under the `### Pangoro Testnet`

AurevoirXavier: Remove or put it in somewhere like `## Install ### nix-pkg`

hujw77: ACK

README.md
```diff
@@ -95,3 +95,10 @@ target/release/drml \
 	--alice \
 	--base-path tests/local-testnet/alice
 ```
+
+## Build
+### NixOS
+```[shell]
```
AsceticBear, line 101: [shell] -> sh? like others.

hujw77: ACK

AsceticBear (approved):

shell.nix
```diff
@@ -0,0 +1,27 @@
+let
+  rust-overlay =
+    import (builtins.fetchGit {
+      url = "https://github.com/oxalica/rust-overlay.git";
+      rev = "84c58400556c1c5fa796cbc3215ba5bbd3bd848f";
+    });
+  nixpkgs = import <nixpkgs> { overlays = [ rust-overlay ]; };
+  rust-nightly = with nixpkgs; (rust-bin.selectLatestNightlyWith (toolchain: toolchain.default.override {
+    extensions = [ "rust-src" ];
```
AurevoirXavier, line 9: I think this is unnecessary

hujw77: I will try it

hujw77: It may be unnecessary for building. But [IDEs like it](https://github.com/paritytech/substrate/commit/6360780056d4cc885b8b066492c03ce7109ebe1d#diff-e53dfbfffe62ae3c0b411b3938ccffa9fb6a2ecc565f55785ef8daa756631a6b)

AurevoirXavier: So, is this an env setup or drml installer?

AurevoirXavier: And if this is an env setup the toolchain that you define here will be overwritten by the rust-toolchain.toml

hujw77: It is a  develop environment for nix user

hujw77: `rust-src` is unnecessary for building.

AurevoirXavier: I've already set up the tools for all the users

https://github.com/darwinia-network/darwinia-common/blob/main/rust-toolchain.toml#L3

hujw77: Not for nix, I think

AurevoirXavier: Once you run any command of the cargo.

The rust-toolchain.toml will overwrite the toolchain.

AurevoirXavier: What's the output of `rustup toolchain list`

hujw77: This is a setup to install cargo and any other related tools.

hujw77: Nix do not install rustup

hujw77: I think [rust-roolchain.toml](https://rust-lang.github.io/rustup/overrides.html) is for user who use rustup manage rust version.
And this is file is for nix user to set up a minimal develop environment.
There are two different ways for different users.

PR #1049, opened 2022-01-17
Remove `KtonAccountBasic` in evm pallet

AsceticBear:


AsceticBear:
I am stuck with the KTON precompile unit tests KTON  after refactoring. I close it, maybe I can find a better way to achieve this later.

PR #1050, opened 2022-01-17
Add DVM To Pangoro

AsceticBear:
Close #1019 , Related PR: https://github.com/ethereum-lists/chains/pull/779

Local test result:

image

PR #1035, opened 2022-01-11
s2s erc20 backing

xiaoch05:
This pr depends on dvm support for backing module.

PR #1053, opened 2022-01-18
Release v2.8.1

AurevoirXavier:
Test with the latest DVM updates.
Then we can make a tag for darwinia-v0.12.0.


HackFisher:
Should we wait for #1050 ?

AurevoirXavier:
> Should we wait for #1050 ?

Nope.
The goal is to test #1031, #1033.

AsceticBear:
> Nope (unnecessary). The goal is to test #1031, #1033. Since #1050 won't affect the downstream (darwinia).

Not necessary. But I suggest including #1050  and #1035  in this release, as these two prs are very close to being merged. One release can save more time to do something else which would be better.

AsceticBear:
Okay. I got it.

AsceticBear (approved):

PR #1062, opened 2022-01-20
Only deposit RING account in genesis

AsceticBear:
The kton precompile unit tests in the DVM module already cover this js tests cases.


AurevoirXavier:
Test failure.

AurevoirXavier:
Don't need KTON tests anymore?

AsceticBear:
> Don't need KTON tests anymore?

See the pull request description I write, The unit tests here: https://github.com/darwinia-network/darwinia-common/blob/main/frame/dvm/src/tests.rs#L1066-L1187

PR #1070, opened 2022-01-25
decode storage with bytes

xiaoch05:


HackFisher (approved):

primitives/contract/src/abi_util.rs
```diff
@@ -58,3 +58,8 @@ pub fn abi_encode_array_bytes32(data: Vec<[u8; 32]>) -> Vec<u8> {
 	let array = data.iter().map(|v| Token::FixedBytes(v.to_vec())).collect();
```
AurevoirXavier, line 58: `data.into_iter().map(|v| Token::FixedBytes(v)).collect();`

primitives/contract/src/abi_util.rs
```diff
@@ -58,3 +58,8 @@ pub fn abi_encode_array_bytes32(data: Vec<[u8; 32]>) -> Vec<u8> {
...
 	ethabi::encode(&[Token::Array(array)])
 }
+
+pub fn abi_encode_array_bytes(data: Vec<Vec<u8>>) -> Vec<u8> {
+	let array = data.iter().map(|v| Token::Bytes(v.to_vec())).collect();
```
AurevoirXavier, line 63: `data.into_iter().map(|v| Token::FixedBytes(v)).collect();`

primitives/contract/src/abi_util.rs
AurevoirXavier: Avoid extra allocation.

primitives/contract/src/abi_util.rs
AurevoirXavier: Avoid extra allocation.

AurevoirXavier (approved):

frame/evm/precompile/contracts/bridge/bsc/src/lib.rs
```diff
@@ -104,16 +101,22 @@ where
...
-						)
-						.map_err(|_| ExitError::Other("verify storage proof failed".into()))?;
-						Ok(storage_value.0.into())
+						);
+						match storage_value {
+							Ok(value) => {
+								return Ok(value.0.clone());
+							}
+							Err(_) => {
+								return Ok(vec![]);
```
AsceticBear, line 113: I am curious about this error here, shouldn't we return Error?  If not, `storage_values?` always OK in my view.

xiaoch05: we should not return error here. If the storage_key is not exist or the storage value is Zero Value, then we get an error here. We should transfer the error to Zero Value because it's multi verify.

AsceticBear: > If the storage_key is not exist or the storage value is Zero Value

You mean the `storage_value`, which is the result of `verify_storage_proof` function?

AsceticBear (approved):

PR #1098, opened 2022-02-23
Fix the potential panic

### LINKED ISSUE #1076: Request unsync block panic in tracing mode
We found a panic issue about tracing in our test network today. The client will panic if we post a `debug_traceBlockByNumber` request to get the block unsync block. This should not happen in my view.

```sh
Thread 'tokio-runtime-worker' panicked at 'called `Option::unwrap()` on a `None` value', /root/Codes/darwinia-network/common/client/dvm/rpc/src/debug.rs:332
```

```sh
// Get the header I want to work with.
let header = client.header(reference_id).unwrap().unwrap();
```

```sh
2022-02-10 05:35:10 The NetworkStart returned as part of `build_network` has been silently dropped
Error: Service(Other("Invalid argument: Column families not opened: col3, col2, col1, col0"))
```

AsceticBear:
Close #1076. Essentially, this is a companion pull request.

### DIFF
--- a/client/dvm/rpc/src/debug.rs
+++ b/client/dvm/rpc/src/debug.rs
@@ -329,7 +329,10 @@ where
 		// Get Blockchain backend
 		let blockchain = backend.blockchain();
 		// Get the header I want to work with.
-		let header = client.header(reference_id).unwrap().unwrap();
+		let header = match client.header(reference_id) {
+			Ok(Some(h)) => h,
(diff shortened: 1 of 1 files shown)


AsceticBear:
@AurevoirXavier PTAR, this is a standalone bug fix pull request.

AurevoirXavier (approved):

PR #1107, opened 2022-03-03
Evm tracing upgrade

AsceticBear:
Upgrade Review Point:
1. Pass `overrides` to debug, trace tasks to adapt more transaction types.
2. Change some `unwrap()` with `expect`, cherry-pick from the upstream repo.
3. Update DebugRuntimeApi version


primitives/evm-trace/events/src/gasometer.rs
```diff
@@ -30,7 +30,6 @@ impl Snapshot {
 	}
 }
 
-#[cfg(feature = "evm-tracing")]
```
HackFisher, line 33: Is this expected?

Remove or just un-comment commented?

primitives/evm-trace/events/src/gasometer.rs
```diff
@@ -30,7 +30,6 @@ impl Snapshot {
 	}
 }
 
-// #[cfg(feature = "evm-tracing")]
```
HackFisher: Re-check, remove or un-comment commented?

primitives/evm-trace/events/src/gasometer.rs
AsceticBear: We can remove it. I have tested the tracing process locally. It works well.

primitives/evm-trace/events/src/gasometer.rs
AsceticBear: See the comments above.

primitives/evm-trace/events/src/gasometer.rs
HackFisher: Just curious, any reason for making this change since enabling this makes the compiler compile less?

client/dvm/rpc/src/debug.rs
```diff
@@ -258,15 +250,19 @@ where
...
-					Some(TracerInput::Blockscout)
-				} else if tracer == "callTracer" {
-					Some(TracerInput::CallTracer)
-				} else {
-					None
-				};
+				const BLOCKSCOUT_JS_CODE_HASH: [u8; 16] =
+					hex_literal::hex!("94d9f08796f91eb13a2e82a6066882f7");
+				const BLOCKSCOUT_JS_CODE_HASH_V2: [u8; 16] =
+					hex_literal::hex!("89db13694675692951673a1e6e18ff02");
```
HackFisher, line 256: What's these magic hashes used for?

AsceticBear: These hashes are used to support multiple callTracers.
 
For example, you can send debug request like
`{ "jsonrpc":"2.0", "id":1, "method":"debug_traceBlockByNumber", "params": ["0xbf", {"tracer": "callTracer"}]}` Or 
https://github.com/PureStake/moonbeam/blob/master/tests/tracing-tests/test-trace.ts#L196-L217.

I do not fully understand the reason they use JSON hash instead of String as param. I guess this is the requirement of [Blockscout](https://blockscout.com/).

primitives/evm-trace/events/src/gasometer.rs
AsceticBear: There may be a slight impact on the size of the binary, the reason for doing this is to be consistent with upstream. I am discussing talks with the moonbeam developers about the possibility of extracting the EVM tracing-related libraries into a separate library so that we can rely on them directly and reduce maintenance costs.

AurevoirXavier (approved):

AsceticBear:
@HackFisher @AurevoirXavier Can we merge this?

HackFisher (approved):

PR #1118, opened 2022-03-08
Fix broken benchmark build

AsceticBear:


HackFisher (approved):

frame/bridge/bsc/src/benchmarking.rs
```diff
@@ -54,25 +52,4 @@ benchmarks! {
...
-
-		Authorities::<T>::put(&initial_authority_set);
-		FinalizedAuthorities::<T>::put(&initial_authority_set);
-		FinalizedCheckpoint::<T>::put(&genesis_header);
-		AuthoritiesOfRound::<T>::insert(
-			&genesis_header.number / T::BscConfiguration::get().epoch_length,
-			(0u32..initial_authority_set.len() as u32).collect::<Vec<u32>>(),
-			);
-		let caller: T::AccountId = T::AccountId::decode(&mut &[0; 32][..]).unwrap_or_default();
-		assert_ok!(<BscBridge<T>>::relay_finalized_epoch_header(
```
HackFisher, line 73: Should we move this to unit tests?

@xiaoch05 Please help check this, are we good to remove it?

xiaoch05: Yes, we can remove it now.

PR #1124, opened 2022-03-10
Add executive param for local test

AsceticBear:
I encountered https://github.com/paritytech/subport/issues/199 this issue again in local test cases. Update the script to avoid others down the same rabbit hole.


AurevoirXavier:
A flag would be better.
`--execution` native will fail the runtime upgrade testing.

AsceticBear:
> A flag would be better. `--execution native` will fail the runtime upgrade testing.

Make sense. Add a command flag, the default execution is `native`~~

AsceticBear:
@AurevoirXavier Do you know what happened to the ci pipe?

AurevoirXavier:
Ignored. This script doesn't affect the runtime/node/test behavior.

tests/local-testnet/deploy.sh
```diff
@@ -49,6 +54,7 @@ do
     --unsafe-ws-external \
     --rpc-methods unsafe \
     --rpc-cors all \
+    --execution $EXECUTION \
```
AurevoirXavier, line 57: What if `$EXECUTION` is empty

AsceticBear: `$EXECUTION` comes from command, If empty, line 16~18, `$EXECUTION=native`

AurevoirXavier: Empty should be default behavior.

AsceticBear: @AurevoirXavier Please review again

PR #1125, opened 2022-03-10
Clean Service Part.1

AurevoirXavier:
Clean the service code which is very helpful for the future darwinia-companion.


AsceticBear:
CI RED

node/cli/src/cli.rs
```diff
@@ -124,8 +120,13 @@ pub enum Subcommand {
 #[derive(Debug, StructOpt)]
 pub struct DvmArgs {
 	/// Enable EVM tracing module on a non-authority node.
-	#[structopt(long, conflicts_with = "validator", require_delimiter = true)]
-	pub ethapi: Vec<EthApiCmd>,
+	#[structopt(
+			long,
+			conflicts_with = "validator",
+			require_delimiter = true,
+			possible_values = &["debug", "trace"]
```
AsceticBear, line 127: According to [possible_values](https://ruststudy.github.io/tao_of_rust_docs/structopt/clap/struct.Arg.html#method.possible_values):
> Specifies a list of possible values for this argument. At runtime, clap verifies that only one of the specified values was used, or fails with an error message.

What we need here is to accept one or two values. such as `--ethapi_debug_targets debug,trace`. I am not sure if `possible_values` can accept two values at the same time.

Another issue is not suggesting changing `ethapi` to `ethapi_debug_targets`, it will break if someone runs a tracing node. If we have to, please update run tracing node crab doc.

node/rpc/src/lib.rs
```diff
@@ -235,12 +204,17 @@ where
 				beefy_commitment_stream,
 				subscription_executor: beefy_subscription_executor,
 			},
-		tracing_requesters,
-		rpc_helper,
+		eth,
```
AsceticBear, line 207: Little thing: Rename to `eth_deps`?

AurevoirXavier: Nope. Follow other fields' spec.

node/cli/src/cli.rs
AurevoirXavier: Passing `debug` and `trace` to `ethabi` make me so confused.
IIRC, there is no debug node on Crab.

AsceticBear (approved):

PR #1128, opened 2022-03-15
Enhance transfer precompile with event

AsceticBear:
The `darwinia-evm` pallet might be replaced with upstream, thus the newly added event goes to the `darwinia-ethereum`. 

## Companioin Note
- PangolinPrecompiles trait bound update from `darwinia_evm::Config` to `darwinia_ethereum::Config`.
- Runtime Ethereum `Event` -> `Event<T>` .


frame/dvm/evm/src/lib.rs
```diff
@@ -129,6 +129,12 @@ pub mod pallet {
 		BalanceDeposit(T::AccountId, H160, U256),
 		/// A withdrawal has been made from a given address. \[sender, address, value\]
 		BalanceWithdraw(T::AccountId, H160, U256),
+		/// Withdraw RING from DVM to darwinia account. \[sender, destination, value\]
+		RingBack(H160, T::AccountId, U256),
+		/// Transfer substrate account KTON to WKTON contract.
+		TransferToWKTON(H160, U256),
```
AurevoirXavier, line 135: WKton

frame/dvm/evm/precompiles/transfer/src/kton.rs
```diff
@@ -108,6 +108,9 @@ impl<T: Config> Kton<T> {
 					}
 				}
 
+				<darwinia_evm::Pallet<T>>::deposit_event(darwinia_evm::Event::TransferToWKTON(
+					caller, value,
+				));
```
xiaoch05, line 113: I think it's better to add this event in contract.

AsceticBear: I have discussed with Echo and XueLei about this solution. Something note:
1. All contract events are wrapped by `Evm::Log(...)` which is hard to catch by application developers, as the Subql can only filter event by module, and cannot filter contract by address. It will cause index overhead for searching transfer precompile related events in Evm::logs.
2. Once WKTON is deployed, it is hard to upgrade.

xiaoch05: We can catch and filter event in contract by thegraph, not `Evm::Log(...)` by subquery. WKTON is an application, it does not belong to EVM pallet.

AsceticBear: @sxlwar Can you explain in more detail why you are unable to use thegraph in your case?

sxlwar: It will be used on the subql side, not thegraph. Subql  can parse substrate events only ,not thegraph events

xiaoch05 (approved):

frame/dvm/ethereum/src/lib.rs
```diff
@@ -302,13 +302,19 @@ pub mod pallet {
...
+	#[pallet::generate_deposit(pub fn deposit_event)]
 	/// Ethereum pallet events.
-	pub enum Event {
+	pub enum Event<T: Config> {
 		/// An ethereum transaction was successfully executed. \[from, to/contract_address, transaction_hash, exit_reason\]
 		Executed(H160, H160, H256, ExitReason),
 		/// DVM Transfer succeeded. \[from, to, value\]
 		DVMTransfer(H160, H160, U256),
+		/// Withdraw RING from DVM to darwinia account. \[sender, destination, value\]
+		RingBack(H160, T::AccountId, U256),
```
AurevoirXavier, line 313: Do we need to rename the `T::AccountId` to `AccountId`

AsceticBear: What's the difference after renaming?

AurevoirXavier: Looks like we don't need to rename the type after metadata v14.

frame/dvm/ethereum/src/lib.rs
```diff
@@ -302,13 +302,19 @@ pub mod pallet {
...
+		/// Transfer substrate account KTON to WKTON contract. \[sender, value\]
+		TransferToWKton(H160, U256),
+		/// Withdraw from WKON contract to substrate account. \[withdraw, value\]
+		WithdrawFromWKton(T::AccountId, U256),
```
HackFisher, line 317: Are these events added on the request of Subscan? I'm not sure whether they are satisfy the requirements. @freehere107

AsceticBear: This request comes from the smart app @sxlwar. I have discussed this with him, it works.

freehere107: It is better to have the event of KTON pallet transfer here

AsceticBear: Native KTON transfer will trigger events from the balances pallet. In the current precompile scenario, this transfer occurs between the substrate native side and the WKTON contract. I do not get the meaning of the event you said above.

frame/dvm/evm/precompiles/transfer/src/kton.rs
```diff
@@ -131,6 +134,10 @@ impl<T: Config> Kton<T> {
 				let target_kton = T::KtonAccountBasic::account_balance(&to);
 				let new_target_kton_balance = target_kton.saturating_add(value);
 				T::KtonAccountBasic::mutate_account_balance(&to, new_target_kton_balance);
```
HackFisher, line 136: Would we reuse `Balance.transfer` instead of `T::KtonAccountBasic::mutate_account_balance` here to trigger balance pallet events, and simply by not introducing new events.

HackFisher:
check non zero for 
```
fn transfer(source: &H160, target: &H160, value: U256) -> Result<(), ExitError> {
```

@freehere107 due the different decimals between native balance and dvm balance, as a result the `ring.transfer` or `kton.transfer` value type u128 can not cover the value in dvm. To make the protocol strictly  correct, we finally decide to have another event for DVM related transfer. We tried to make the interfaces as simple/less as possible, but unfortunately, we must add another event/interface after discussion.

So, in subscan, you will have to track two events for ring/kton, `balance.transfer` and DVMTransfer.

frame/dvm/ethereum/src/lib.rs
```diff
@@ -302,13 +302,13 @@ pub mod pallet {
...
+	#[pallet::generate_deposit(pub fn deposit_event)]
 	/// Ethereum pallet events.
-	pub enum Event {
+	pub enum Event<T: Config> {
 		/// An ethereum transaction was successfully executed. \[from, to/contract_address, transaction_hash, exit_reason\]
 		Executed(H160, H160, H256, ExitReason),
-		/// DVM Transfer succeeded. \[from, to, value\]
-		DVMTransfer(H160, H160, U256),
+		/// DVM Transfer succeeded. \[AddrWrapper(from), AddrWrapper(to), value\]
+		DVMTransfer(AddrWrapper<T::AccountId>, AddrWrapper<T::AccountId>, U256),
```
HackFisher, line 311: Why not directly use `T::AccountId` ? `AddrWrapper` introduce extra enum type here and unnecessary match in implementations.

HackFisher (changes requested):

frame/dvm/ethereum/src/account_basic.rs
```diff
@@ -118,52 +118,63 @@ where
...
 	/// Get the account basic in EVM format.
-	fn account_basic(address: &H160) -> EVMAccount {
-		let account_id = <T as darwinia_evm::Config>::IntoAccountId::into_account_id(*address);
+	fn account_basic(addr: AddrWrapper<T::AccountId>) -> EVMAccount {
+		let account_id = match addr {
+			AddrWrapper::ETH(addr) => {
+				<T as darwinia_evm::Config>::IntoAccountId::into_account_id(addr)
+			}
+			AddrWrapper::Substrate(account_id) => account_id,
+		};
```
HackFisher, line 127: Prefer to add two method if required here instead of match switch here. Let's keep it simple.

frame/dvm/ethereum/src/account_basic.rs
```diff
@@ -118,52 +118,63 @@ where
...
 	}
 
 	/// Mutate account balance.
-	fn mutate_account_balance(account_id: &T::AccountId, new_balance: U256) {
+	fn mutate_account_basic_balance(addr: AddrWrapper<T::AccountId>, new_balance: U256) {
+		let account_id = match addr {
+			AddrWrapper::ETH(addr) => {
+				<T as darwinia_evm::Config>::IntoAccountId::into_account_id(addr)
+			}
+			AddrWrapper::Substrate(account_id) => account_id,
```
HackFisher, line 172: Use two methods instead of type switch.

frame/dvm/ethereum/src/lib.rs
AsceticBear: Actually `AddrWrapper` make `account_basic` implementation much easier. We do not need to split some functions into two versions, `evm_transfer` and `substrate_transfer` anymore. And for the application developer, this event is much more readable in my view.

frame/dvm/evm/src/lib.rs
```diff
@@ -476,18 +485,24 @@ pub trait EnsureAddressOrigin<OuterOrigin> {
 	) -> Result<Self::Success, OuterOrigin>;
 }
 
+#[derive(Debug, Clone, Encode, Decode, Eq, PartialEq, Copy, TypeInfo)]
+pub enum AddrWrapper<AccountId> {
```
AurevoirXavier, line 489: Looks like duplicate with `sp_runtime::MultiAddress`

frame/dvm/ethereum/src/lib.rs
HackFisher: First, `AddrWrapper ` will make the user of this event pay extra attention to decode it, it break consistency of subscan current account interpreter, subscan will need to pay more attention for the first/byte encoded by the enum.

Second, to be frankly, I think using two method here is much more simpler and easy to read here, I don't need to knwo what's the concept of AddrWrapper and the `match switch` bad pattern.

AsceticBear: > First, AddrWrapper will make the user of this event pay extra attention to decode it, it break consistency of subscan current account interpreter/render, subscan will need to pay more attention for the first byte encoded by the enum.

After V14, no JSON type is needed, the subscan side can decode this struct directly without extra effort. If using `AddrWrapper`, they don't need to do conversation from AccountId to H160. Otherwise, they need this convert.

AsceticBear:
@HackFisher Please take a review again.

HackFisher (approved):

AsceticBear:
@AurevoirXavier Do you know why the CI always skipped some check halfway?

AurevoirXavier:
> @AurevoirXavier Do you know why the CI always skipped some check halfway?

I'm adjusting the susbtrate#darwinia-v0.12.1.
I'll fix this later.

AurevoirXavier:
CI failed. @AsceticBear

frame/dvm/evm/precompiles/transfer/Cargo.toml
```diff
@@ -29,6 +30,24 @@ sp-std        = { default-features = false, git = "https://github.com/darwinia-n
 # frontier
 fp-evm = { default-features = false, git = "https://github.com/darwinia-network/frontier", branch = "darwinia-v0.12.1-dev" }
 
+[dev-dependencies]
+# crates.io
+ethereum     = { version = "0.11.1", default-features = false, features = ["with-codec"] }
```
AurevoirXavier, line 35: It seems that we don't need the `default-features = false`

AurevoirXavier (approved):

PR #1133, opened 2022-03-18
Simplify return error way

AsceticBear:
No logic change. This is an easy one to review.


frame/dvm/evm/precompiles/utils/src/lib.rs
```diff
@@ -46,3 +46,9 @@ impl<'a> DvmInputParser<'a> {
 		})
 	}
 }
+
+pub fn ret_err(err_msg: &'static str) -> PrecompileFailure {
```
AurevoirXavier, line 50: This name is ambiguous. 
`custom_precompile_err(custom_err_msg: &str)`.

HackFisher (approved):

PR #1138, opened 2022-03-22
Add precompile function checker

AsceticBear:


xiaoch05 (approved):

frame/dvm/evm/precompiles/bridge/ethereum/src/lib.rs
```diff
@@ -50,11 +52,16 @@ where
...
-		_is_static: bool,
+		context: &Context,
+		is_static: bool,
 	) -> PrecompileResult {
 		let dvm_parser = DvmInputParser::new(&input)?;
-		let output = match Action::from_u32(dvm_parser.selector)? {
+		let action = Action::from_u32(dvm_parser.selector)?;
+
+		// Check state modifiers
+		check_state_modifier(context, is_static, StateMutability::View)?;
```
HackFisher, line 62: Not sure this `StateMutability` is `View`, @xiaoch05 .

Besides, this is better to be checked in `Action` level, different actions might have different `StateMutability `

xiaoch05: I don't know if it makes sense to add this `check_state_modifier`, since all the calls happen inside the dispatch precompile contract.

HackFisher (approved):

PR #1156, opened 2022-04-02
Adjust Primitives

AurevoirXavier:
I didn't see too much usage of dp-fee.
And it should a part of fee-market.
Like `staking::Ledger`, `democracy::Vote` which were defined in the pallet.

Also, we can easily integrate it into parachain through a single pallet.


AsceticBear:
RED CI

frame/fee-market/src/lib.rs
```diff
@@ -246,13 +245,14 @@ pub mod pallet {
 
 			// Increase the locked collateral
 			if new_collateral >= Self::relayer(&who).collateral {
-				let _ = T::RingCurrency::extend_lock(
+				T::RingCurrency::set_lock(
```
AsceticBear, line 248: I need some time to check the change here, please fix ci firstly.

HackFisher (approved):
LGTM

frame/fee-market/src/lib.rs
AurevoirXavier: If you have already compared the value. Just set a new lock here and you can get rid of this error handling.

AsceticBear: I see the concrete implementation difference in this.  LGTM, just waiting for the ci.

AsceticBear (approved):

PR #1158, opened 2022-04-02
Update Substrate and Fix `darwinia-fee-market`

AurevoirXavier:
`AccoundId` is not more defaultable.
Hence we need to change `ValueQuery` to `OptionQuery`.
Please fix the code logic. @AsceticBear

### DIFF
--- a/frame/fee-market/src/lib.rs
+++ b/frame/fee-market/src/lib.rs
@@ -37,7 +37,7 @@ use bp_messages::{LaneId, MessageNonce};
 use frame_support::{
 	ensure,
 	pallet_prelude::*,
-	traits::{Currency, Get, LockIdentifier, WithdrawReasons},
+	traits::{Currency, Get, LockIdentifier, LockableCurrency, WithdrawReasons},
 	transactional, PalletId,
 };
 use frame_system::{ensure_signed, pallet_prelude::*};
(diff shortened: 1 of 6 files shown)


frame/fee-market/src/lib.rs
```diff
@@ -138,7 +134,7 @@ pub mod pallet {
 		Blake2_128Concat,
 		T::AccountId,
 		Relayer<T::AccountId, RingBalance<T>>,
-		ValueQuery,
+		OptionQuery,
```
HackFisher, line 137: Do we need to migrate storage if we change to OptionQuery?

I remember that it will not affect decoding, right?

HackFisher (commented):
We changed the substrate deps in this branch, but only in the scope of fee-market pallet.

Only this pallet compiles and other will fail right? Is that sufficient for the requirement of Crab/Pangolin parachain?

frame/fee-market/src/lib.rs
AurevoirXavier: No need to.

AurevoirXavier:
> We changed the substrate deps in this branch, but only in the scope of fee-market pallet.
> 
> Only this pallet compiles and other will fail right? Is that sufficient for the requirement of Crab/Pangolin parachain?

Yes. The parachains only need this pallet.

AsceticBear:
> AccoundId is not more defaultable.

Can you find the upstream substrate pr related to this change? @AurevoirXavier

AsceticBear:
Note: I upgrade the `pallet-bridge-messages` with https://github.com/darwinia-network/darwinia-bridges-substrate/commit/07bbb7427ffcf12ef9c7ce66bc2f64fb08525b08 to pass the units tests.

HackFisher (approved):

PR #1160, opened 2022-04-02
Release v2.8.7

AurevoirXavier:


HackFisher (approved):
The substrate deps is not changed, and there are no other migrations since last Pangolin release, so no migrations?

LGTM

AsceticBear (approved):

PR #1173, opened 2022-04-13
Fix gas issues

AsceticBear:
1. Add a gas-meter for calculation precompile used gas.
2. Update Ethereum `BlockGasLimit` according to `BlockWeightLimit`.
3. Update darwinia-evm benchmark test result and `GasWeightMapping` config.
4. Add test about the max gas limit of normal ethereum transact extrinsic.

# Companion Tips

1. Use `impl_self_contained_call` macro to reduce duplication in runtime.
5. Update `FixedGasWeightMapping` in darwinia-evm pallet config
6. Update `BlockGasLimit` in darwinia-evm pallet config


AsceticBear:
@HackFisher Please take a review about this

node/runtime/common/src/lib.rs
```diff
@@ -133,3 +133,6 @@ frame_support::parameter_types! {
 		.max
 		.get(DispatchClass::Normal);
 }
+
+// Accounding to the evm gas benchmark, 1 gas ~= 40_000 weight.
+pub const WEIGHT_PER_GAS: u64 = 40_000;
```
HackFisher, line 138: Need review and double check.

tests/dvm/darwinia/tests/config.ts
```diff
@@ -6,3 +6,6 @@ export const config = {
 	gas: 4000000,
 	jsontestAddress: "",
 };
+
+// EXTRINSIC_GAS_LIMIT = [BLOCK_GAS_LIMIT - BLOCK_GAS_LIMIT * (NORMAL_DISPATCH_RATIO - AVERAGE_ON_INITIALIZE_RATIO) - EXTRINSIC_BASE_Weight] / WEIGHT_PER_GAS  = (1_000_000_000_000 * 2 * (0.75-0.025) - 125_000_000) / 40_000
+export const EXTRINSIC_GAS_LIMIT = 36246875;
```
HackFisher, line 11: Need review and double check.

HackFisher (approved):

node/runtime/common/src/lib.rs
AsceticBear: Yeah, this is a critical variable. The value I choose considering the current EVM benchmark file and other project configurations. We can upgrade to the pangolin test network firstly and do more tests about this. If necessary, update later.

tests/dvm/darwinia/tests/config.ts
AsceticBear: This constant is related to the `WEIGHT_PER_GAS`, If `WEIGHT_PER_GAS` is right, this value is right too. There is a test case designed for the calculation process.

PR #1174, opened 2022-04-15
Revamp Bridge Primitives

AurevoirXavier:


node/runtime/pangoro/src/pallets/bridge_dispatch.rs
```diff
@@ -14,7 +14,7 @@ impl Config<WithPangolinDispatch> for Runtime {
 	type CallFilter = Everything;
 	type EncodedCall = bm_pangolin::FromPangolinEncodedCall;
 	type SourceChainAccountId = bp_pangolin::AccountId;
-	type TargetChainAccountPublic = AccountPublic;
-	type TargetChainSignature = Signature;
+	type TargetChainAccountPublic = bp_pangoro::AccountPublic;
+	type TargetChainSignature = bp_pangoro::Signature;
```
AsceticBear, line 18: According to the type name `TargetChain...`, I guess this should be `bp_pangolin`.

AurevoirXavier: https://github.com/darwinia-network/darwinia-bridges-substrate/blob/main/bin/millau/runtime/src/lib.rs#L225-L235

PR #1182, opened 2022-04-19
Add build_message_payload(...) method to s2s precompile

wuminzhe:
if you want to do a `smart chain > pallet` remote dispatch_call, the dispatch_call data need to be wrapped with some other information.  The wrapped data is the `message payload`. This is a scale codec process which is hard to do in solidity. So, I add a  `build_message_payload` helper method to the s2s precompile.

### DIFF
--- a/frame/dvm/evm/precompiles/bridge/s2s/src/lib.rs
+++ b/frame/dvm/evm/precompiles/bridge/s2s/src/lib.rs
@@ -32,7 +32,7 @@ use darwinia_support::{
 };
 use dp_contract::{
 	abi_util::{abi_decode_bytes4, abi_encode_bytes, abi_encode_u64},
-	mapping_token_factory::s2s::{S2sRemoteUnlockInfo, S2sSendMessageParams},
+	mapping_token_factory::s2s::{S2sRemoteUnlockInfo, S2sSendMessageParams, S2sMessagePayloadInfo},
 };
 use dp_s2s::{CallParams, CreatePayload};
 // --- paritytech ---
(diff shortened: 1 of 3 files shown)


AurevoirXavier:
Why the target branch is darwinia-v0.12.1?

wuminzhe:
Which branch will be better?

frame/dvm/evm/precompiles/bridge/s2s/src/lib.rs
```diff
@@ -159,4 +164,22 @@ where
...
 	}
+
+	fn build_message_payload(
+		dvm_parser: &DvmInputParser,
+		caller: H160,
+	) -> Result<Vec<u8>, PrecompileFailure> {
+		let info = S2sMessagePayloadInfo::abi_decode(dvm_parser.input)
+			.map_err(|_| custom_precompile_err("abi_decode message payload info failed"))?;
+
+		let payload = <T as from_substrate_issuing::Config>::ToPangoroMessagePayload {
```
AsceticBear, line 175:
```suggestion
		let payload = <T as from_substrate_issuing::Config>::ToPangoroMessagePayload {
```

This is a general precompile used in s2s message delivery. you cannot introduce things like `ToPangoroMessagePayload`. `ToPangoroMessagePayload` is a chain runtime concept.  I suggest refactoring the `encode_send_message_dispatch_call` S2sSendMessageParams to include necessary field for decode payload, like spec_version, weight. @xiaoch05

wuminzhe: that's right. this method is the generic version of `encode_send_message_dispatch_call`. I will change ToPangoroMessagePayload to generic payload.

AsceticBear:
Prefer #1186

PR #1186, opened 2022-04-19
S2s precompile support general call encode

AsceticBear:


frame/dvm/evm/precompiles/bridge/s2s/src/lib.rs
```diff
@@ -148,6 +156,23 @@ where
 		Ok(abi_encode_bytes(payload.encode().as_slice()))
 	}
 
+	fn create_opaque_call_payload(
```
wuminzhe, line 159: maybe, this name is not friendly to DApp developer. It will be exposed as a SDK API.
It returns the final cross-chain message payload, and it doesn't create anything but assemble incoming stuff, so I prefer `build_message_payload`.

wuminzhe: or just `build_message`

wuminzhe:
```solidity
contract PangolinSmart2PangoroDemo is MappingTokenAddress {
    event PrintMessage(bytes msg);

    /// This can be run in
    ///   1. "Pangolin Smart Chain" to remote call Pangoro's System.remark
    ///   2. "Crab Smart Chain" to remote call Darwinia's System.remark
    function remoteCallSystemRemark() public payable {
        ///////////////////////////////////////////
        // Prepare a message for Target Chain 
        ///////////////////////////////////////////
        // bytes of System.remark("hellow world")
        bytes memory systemRemark = "0x00012c68656c6c6f20776f726c64"; 
        // build the cross chain message
        bytes memory systemRemarkMessage = buildMessage(28070, 10000000, systemRemark);

        ///////////////////////////////////////////
        // send the message from Source Chain
        ///////////////////////////////////////////
        sendMessage(43, 0, systemRemarkMessage);
    }

    /// in SDK
    function sendMessage(uint32 palletIndex, bytes4 laneId, bytes memory messagePayload) public payable {
        uint256 fee = msg.value/(10**9); // the pricision in contract is 18, and in pallet is 9, transform the fee value
        bytes memory sendMessageCall = SubToSubBridge(DISPATCH_ENCODER).encode_send_message_dispatch_call(
            palletIndex, // pallet index of source chain.
            laneId, // laneid
            messagePayload,
            fee
        );
        
        (bool success, ) = DISPATCH.call(sendMessageCall);
        require(success, "send message failed");
    }

    /// in SDK
    function buildMessage(uint32 specVersion, uint64 weight, bytes memory callData) public payable returns (bytes memory) {
        // message payload for call bytes.
        bytes memory messagePayload = SubToSubBridge(DISPATCH_ENCODER).build_message_payload(
            specVersion, // specVersion, system.version
            weight, // weight, how to get
            callData // call bytes
        );
        return messagePayload;
    }
}
```

frame/dvm/evm/precompiles/bridge/s2s/src/lib.rs
AsceticBear: `build_message` is too simple in this case. `assemble incoming stuff,` also means create in my view. So I prefer the current name style. Wait and see what others review opinions?

wuminzhe: see my solidity example below. the SDK will provide a pair methods: `sendMessage` and `buildMessage`. And `buildMessage` is just a wrapper for the precompile method. So, the similar name will be better.

wuminzhe: `create` means you create a new buisness thing.
`build` means the input and the output are the same business thing. only the format or encoding is different.

wuminzhe (approved):

primitives/contract/src/mapping_token_factory/s2s.rs
```diff
@@ -155,6 +155,32 @@ impl S2sRemoteUnlockInfo {
 	}
 }
 
+#[derive(Debug, PartialEq, Eq)]
+pub struct S2sOpaqueCallParams {
```
xiaoch05, line 159: this is not belong to mapping_token_factory
It's better to move it outside

xiaoch05:
Each file here corresponds to each file in the contract，we should not rename them. Just add another enocde file.

xiaoch05:
> > Each file here corresponds to each file in the contract， we should not rename them. Just add another encode file. @AsceticBear
> 
> I have thought about this. I don't know where to put this encode function and creating a new file for it seems ugly to me. Considering your review comment that the `S2sOpaqueCallParams` does not belong to the mapping-token-factor, this makes sense. After going over the whole file structure, the `S2sOpaqueCallParams` is also part of the precompile actually, all structs that need to decode go to precompile. Maybe I should put precompile under mapping-token-factor?

The new added encoder has no relationship with mapping-token-factory. It should not be put in that directory.
And this directory structor is from contracts, so do not move or rename the file.

wuminzhe:
`encode_send_message_dispatch_call`'s params may be : 
`pallet_index`,
`lane_id`,
`payload`,
`instance`

The last one is changed from the fee to the bridge instance just like the rpc api param.
The fee will be fetched in the `encode_send_message_dispatch_call` precompile.

HackFisher:
Will be replace by the approach of using solidity library:

https://github.com/darwinia-network/darwinia-bridges-sol/issues/154

PR #1187, opened 2022-04-19
Use Parity `LockableCurrency`

AurevoirXavier:


AurevoirXavier:
Please review the rewritten logic (without struct changes) first.
Then decide if we should make some migration.

AsceticBear:
What is the next plan for this pull request? I noticed you retain the `OldBalanceLock` in the balance module, does that mean no migration about balance lock storage in this pull request? (It's not a easy one for me to review, hh)

frame/bridge/relay-authorities/src/lib.rs
```diff
@@ -305,7 +304,7 @@ decl_module! {
 				<Error<T, I>>::AuthorityAE
 			);
 			ensure!(
-				<RingCurrency<T, I>>::usable_balance(&account_id) > stake,
+				<RingCurrency<T, I>>::free_balance(&account_id) > stake,
```
AsceticBear, line 307: Is this expected? @HackFisher Please check it later.

frame/support/src/structs.rs
```diff
@@ -52,69 +53,37 @@ where
...
-impl<Balance, Moment> BalanceLock<Balance, Moment>
+impl<Balance, Moment> OldBalanceLock<Balance, Moment>
 where
 	Balance: Copy + PartialOrd + AtLeast32BitUnsigned,
 	Moment: Copy + PartialOrd,
 {
 	// For performance, we don't need the `at` in some cases
 	// Only use for tests to avoid write a lot of matches in tests
-	pub fn locked_amount(&self, at: Option<Moment>) -> Balance {
+	pub fn locked_amount(&self) -> Balance {
```
AsceticBear, line 83: Remember to update the doc, no `at` any more.

frame/claims/src/lib.rs
```diff
@@ -352,7 +351,8 @@ impl<T: Config> Module<T> {
 
 	fn pot<C: LockableCurrency<T::AccountId>>() -> C::Balance {
 		// Already lock minimal balance in the account, no need to worry about to be 0.
-		C::usable_balance(&Self::account_id())
```
AsceticBear, line 355: Looks like the `usable_balance` is only used for balance rpc?

AurevoirXavier: Almost.
And I think we can completely remove this later.

AurevoirXavier:
> What is the next plan for this pull request? I noticed you retain the `OldBalanceLock` in the balance module, does that mean no migration about balance lock storage in this pull request? (It's not a easy one for me to review, hh)

Depending on the migration performance testing.

HackFisher (approved):

HackFisher (approved):

HackFisher:
Let's do the OldBalanceLock migration first, and leave other migrations to later runtime upgrade.

PR #1192, opened 2022-04-22
add parachain backing

xiaoch05:


frame/wormhole/issuing/s2s/src/lib.rs
```diff
@@ -165,7 +165,7 @@ pub mod pallet {
 			origin: OriginFor<T>,
 			token_address: H160,
 			amount: U256,
-			recipient: H160,
+			recipient: Vec<u8>,
```
HackFisher, line 168: Are you try to generalize this?

If not that easy, I would like to keep it simple since these pallets will be reimplemented using message SDK any way, It does not deserve more time on this.

Just straight forward and keep simple, we do not have to reuse existing pallet. 

cc @xiaoch05 @wuminzhe

HackFisher (commented):
Any progress or ETA on this?

xiaoch05:
> Any progress or ETA on this?

The test has been finished.
Will ready for review after some code clean.

xiaoch05:
use https://github.com/darwinia-network/darwinia-common/pull/1243

PR #1204, opened 2022-04-28
Add state storage precompile

AsceticBear:
An example:
```sol
// SPDX-License-Identifier: MIT
pragma solidity >=0.8.10;

interface IStateStorage {
    function state_storage(bytes memory) external view returns (bytes memory);
}

contract TestStateStorage {
    function state_storage(bytes memory key) public view returns (bytes memory) {
        address addr = 0x0000000000000000000000000000000000000001;
        return IStateStorage(addr).state_storage(key);
    }
}
```
Note: 
1. The value comes from `state_storage(xxx)` is ethabi encoded. @wuminzhe 
2. [Substrate Storage Key](https://docs.substrate.io/v3/advanced/storage/)  A very helpful doc when you want to construct a storage key.


AurevoirXavier:
Who can review this?

AurevoirXavier:
Who can review this?

AsceticBear:
@HackFisher Please help take a review

HackFisher:
I'm wondering wether we should make it more generalized, replaced by a reading substrate storage precompile, passing in the the storage key and output the storage value bytes.

Since precompile is the interface for smart contract developer, it deserves more consideration in this aspect.

What's your ideas on this? @hujw77 @wuminzhe @xiaoch05

xiaoch05:
> I'm wondering wether we should make it more generalized, replaced by a reading substrate storage precompile, passing in the the storage key and output the storage value bytes.
> 
> Since precompile will be long live interface for smart contract developer, it deserves more consideration in this aspect.
> 
> What's your ideas on this? @hujw77 @wuminzhe @xiaoch05

I also agree with this generic approach, and contract developers are generally capable of converting such generic bytes to the type they need.
If we expose too many special interfaces in the Precompile, it will be difficult to remove them in the future once they are used inside the contract.

hujw77:
> I'm wondering wether we should make it more generalized, replaced by a reading substrate storage precompile, passing in the the storage key and output the storage value bytes.
> 
> Since precompile will be long live interface for smart contract developer, it deserves more consideration in this aspect.
> 
> What's your ideas on this? @hujw77 @wuminzhe @xiaoch05

The storage is public at off-chain view. But on-chain it is also have [State Variable Visibility](https://docs.soliditylang.org/en/v0.8.13/contracts.html#visibility-and-getters).
If the generic function make all state varible public, it is not safety.
Use interface to access other contract/pallet state is better.

hujw77:
> > I'm wondering wether we should make it more generalized, replaced by a reading substrate storage precompile, passing in the the storage key and output the storage value bytes.
> > Since precompile will be long live interface for smart contract developer, it deserves more consideration in this aspect.
> > What's your ideas on this? @hujw77 @wuminzhe @xiaoch05
> 
> The storage is public at off-chain view. But on-chain it is also have [State Variable Visibility](https://docs.soliditylang.org/en/v0.8.13/contracts.html#visibility-and-getters). If the generic function make all state varible public, it is not safety. Use interface to access other contract/pallet state is better.

Another idea, If there is a generic precompile / way to call substate public function.

AsceticBear:
> The storage is public at off-chain view. But on-chain it is also have [State Variable Visibility](https://docs.soliditylang.org/en/v0.8.13/contracts.html#visibility-and-getters).

The state visibility referred to seems only designed to restrict the state access from the contract view. Even if you added the `private`, the private storage can be readable using more low-level interface.

xiaoch05:
> > I'm wondering wether we should make it more generalized, replaced by a reading substrate storage precompile, passing in the the storage key and output the storage value bytes.
> > Since precompile will be long live interface for smart contract developer, it deserves more consideration in this aspect.
> > What's your ideas on this? @hujw77 @wuminzhe @xiaoch05
> 
> The storage is public at off-chain view. But on-chain it is also have [State Variable Visibility](https://docs.soliditylang.org/en/v0.8.13/contracts.html#visibility-and-getters). If the generic function make all state varible public, it is not safety. Use interface to access other contract/pallet state is better.

Where does this insecurity come from? 
All this storage externally accessible, and can our interfaces all be read-only? There should be no ability to modify this storage inside the contract.

hujw77:
> > > I'm wondering wether we should make it more generalized, replaced by a reading substrate storage precompile, passing in the the storage key and output the storage value bytes.
> > > Since precompile will be long live interface for smart contract developer, it deserves more consideration in this aspect.
> > > What's your ideas on this? @hujw77 @wuminzhe @xiaoch05
> > 
> > 
> > The storage is public at off-chain view. But on-chain it is also have [State Variable Visibility](https://docs.soliditylang.org/en/v0.8.13/contracts.html#visibility-and-getters). If the generic function make all state varible public, it is not safety. Use interface to access other contract/pallet state is better.
> 
> Where does this insecurity come from? All this storage externally accessible, and can our interfaces all be read-only? There should be no ability to modify this storage inside the contract.

```solidity
contract A {
     uint private a;
}
```
There is no way to read the `a` value on evm(not including substrate side using low-level method). If we make all state open, it is a break change.

HackFisher:
To be compatible with evm solidity's visibility design, storage values from EVM and Ethereum pallet can be excluded from the precompile, by filtering the storage key using the pallet's storage prefix.

AsceticBear:
@HackFisher @xiaoch05  Please take a review.

node/runtime/pangolin/src/pallets/evm.rs
```diff
@@ -81,6 +82,13 @@ impl LatestMessageNoncer for ToPangoroMessageSender {
 	}
 }
 
+pub struct StorageFilter;
+impl StorageFilterT for StorageFilter {
+	fn allow(prefix: &[u8]) -> bool {
+		prefix != Twox128::hash(b"EVM") && prefix != Twox128::hash(b"Ethereum")
```
xiaoch05, line 88: the prefix is defined as "dvm", not like this EVM?
https://github.com/darwinia-network/darwinia-common/blob/main/frame/support/src/evm.rs#L75
and why we need put this Filter into runtime? I think it's a primitives function.

frame/dvm/evm/precompiles/utils/Cargo.toml
```diff
@@ -11,16 +11,22 @@ version     = "2.8.11"
...
 [dependencies]
 # crates.io
-ethabi = { version = "15.0", default-features = false }
-evm    = { version = "0.33.1", default-features = false, features = ["with-codec"] }
+ethabi       = { version = "15.0", default-features = false }
+ethereum     = { version = "0.11.1", features = ["with-codec"], optional = true }
+evm          = { version = "0.33.1", default-features = false, features = ["with-codec"] }
+libsecp256k1 = { version = "0.5", features = ["static-context", "hmac"], default-features = false, optional = true}
+rlp          = { version = "0.5", optional = true}
+sha3         = { version = "0.9", default-features = false, optional = true}
```
xiaoch05, line 19: should some of the above packages be moved to the dev-dependencies?

frame/dvm/evm/precompiles/state-storage/src/lib.rs
```diff
@@ -0,0 +1,95 @@
...
+				helper.record_gas(1, 0)?;
+
+				frame_support::storage::unhashed::get_raw(key)
+			},
+		};
+
+		Ok(PrecompileOutput {
+			exit_status: ExitSucceed::Returned,
+			cost: helper.used_gas(),
+			output: abi_encode_bytes(&output.unwrap_or_default()),
```
xiaoch05, line 91: This default value may be mistaken by the application as a valid value.

frame/dvm/evm/precompiles/state-storage/src/lib.rs
```diff
@@ -0,0 +1,95 @@
...
+		let output = match action {
+			Action::StateGetStorage => {
+				let tokens = ethabi::decode(&[ParamType::Bytes], data)
+					.map_err(|_| helper.revert("ethabi decoded error"))?;
+				let key = match &tokens[0] {
+					Token::Bytes(bytes) => bytes,
+					_ => return Err(helper.revert("The input param error")),
+				};
+
+				if key.len() < 16 || !F::allow(&key[0..16]) {
```
xiaoch05, line 77: What does this 16 stand for? it's better to define a constant value here.

node/runtime/pangolin/src/pallets/evm.rs
AsceticBear: > the prefix is defined as "dvm", not like this EVM?

Yes, We use the module name when calculating the storage key. After discussion with Echo and Denny,  we want to filter any state storage query related to EVM and Ethereum module to keep compatible with evm design.

frame/dvm/evm/precompiles/state-storage/src/lib.rs
```diff
@@ -0,0 +1,95 @@
...
+		let mut helper = PrecompileHelper::<T>::new(input, target_gas);
+		let (selector, data) = helper.split_input()?;
+		let action = Action::from_u32(selector)?;
+
+		// Check state modifiers
+		helper.check_state_modifier(context, is_static, StateMutability::View)?;
+
+		let output = match action {
+			Action::StateGetStorage => {
+				let tokens = ethabi::decode(&[ParamType::Bytes], data)
```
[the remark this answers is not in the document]
AsceticBear: I don't get your point here, I use `ParamType::Bytes` here.

frame/dvm/evm/precompiles/state-storage/src/lib.rs
AsceticBear: 16 = `Twox128::hash(b"EVM")` length. You are right. Give it a constant value is better.

frame/dvm/evm/precompiles/state-storage/src/lib.rs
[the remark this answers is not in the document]
xiaoch05: the length of the key is 16+16=32?
So we can use `ParamType::FixedBytes(32)` here

frame/dvm/evm/precompiles/state-storage/Cargo.toml
```diff
@@ -0,0 +1,65 @@
...
+authors     = ["Darwinia Network <hello@darwinia.network>"]
+description = "State storage precompiles for EVM pallet."
+edition     = "2021"
+homepage    = "https://darwinia.network"
+license     = "GPL-3.0"
+name        = "darwinia-evm-precompile-state-storage"
+readme      = "README.md"
+repository  = "https://github.com/darwinia-network/darwinia-common"
+version     = "2.8.11"
+
```
AurevoirXavier, line 11: extra empty line

frame/dvm/evm/precompiles/state-storage/src/lib.rs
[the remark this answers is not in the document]
AsceticBear: > the length of the key is 16+16=32?

No, What if the storage item is a map, and the key's length is larger than 32 bytes. For the simplest `StorageValue` type, the key is 32 bytes.

frame/dvm/evm/precompiles/utils/Cargo.toml
AsceticBear: I checked it again, some deps are under `testing` feature control.

frame/dvm/evm/precompiles/state-storage/src/lib.rs
```diff
@@ -0,0 +1,97 @@
...
+use evm::ExitRevert;
+// --- darwinia-network ---
+use darwinia_evm_precompile_utils::PrecompileHelper;
+use dp_contract::abi_util::abi_encode_bytes;
+// --- paritytech ---
+use fp_evm::{
+	Context, ExitSucceed, Precompile, PrecompileFailure, PrecompileOutput, PrecompileResult,
+};
+
+const PALLET_PREFIX: usize = 16;
```
HackFisher, line 37:
```suggestion
const PALLET_PREFIX_Length: usize = 16;
```

AsceticBear: Done

HackFisher (approved):

PR #1200, opened 2022-04-27
Replace Fee Market

AsceticBear:
## TODO
- Companion for darwinia

PR #1208, opened 2022-05-06
Release `v2.8.11`

AurevoirXavier:
Apply #1202. For S2S bridge testing.


HackFisher:
> Suggest waiting for a while and let #1204 #1200 in, then the sdk team can do something with fee-market precompile.

OK for #1200, but #1204 might require more review and discussion, how about delay 1204 to next runtime version?

AsceticBear:
> how about delay 1204 to next runtime version?

Sure. @AurevoirXavier Please fix the conflicts files

HackFisher (changes requested):
Did we miss BalanceLock migrations in this runtime upgrade?

PR #1209, opened 2022-05-07
Shorten Feemarket Slot 

AsceticBear:
Companion for darwinia https://github.com/darwinia-network/darwinia/pull/875

### DIFF
--- a/frame/balances/src/migration.rs
+++ b/frame/balances/src/migration.rs
@@ -10,7 +10,7 @@ pub use try_runtime::*;
 // --- darwinia-network ---
 use crate::*;
 
-pub fn migrate_lock<T, I>()
+pub fn migrate<T, I>()
 where
 	T: Config<I>,
 	I: 'static,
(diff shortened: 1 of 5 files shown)


frame/balances/src/migration.rs
```diff
@@ -10,7 +10,7 @@ pub use try_runtime::*;
 // --- darwinia-network ---
 use crate::*;
 
-pub fn migrate<T, I>()
+pub fn migrate_lock<T, I>()
```
AurevoirXavier, line 13: Prefer to keep this.
And I don't need to change on darwinia side.

AsceticBear: I have thought about this so I just rename it. We might have other migrations in the future, it's better to keep the migration history?

AurevoirXavier: I think it isn't the time to do that.
If a lot of teams are using our pallet, I would consider that.

PR #1214, opened 2022-05-13
Update dispatch returned error

AsceticBear:


AurevoirXavier:
I think the error will be encoded automatically while throwing it out.
Did you test that?

frame/dvm/evm/precompiles/dispatch/src/lib.rs
```diff
@@ -81,7 +81,7 @@ where
 					logs: Default::default(),
 				})
 			},
-			Err(_) => Err(helper.revert("dispatch execution failed")),
+			Err(e) => Err(helper.revert(Encode::encode(&e))),
```
wuminzhe, line 84: I want to confirm that the result returned to solidity is not abi encoded.

HackFisher:
Ready for review, anything concern else?

AsceticBear:
> Ready for review? anything concern else?

Yeah, there is something we have discussed this afternoon, refer to https://github.com/darwinia-network/darwinia-common/issues/1217 . @wuminzhe  I've preferred to implement what was discussed in this issue in another pr, so please take a review about this first.

HackFisher:
> > Ready for review? anything concern else?
> 
> Yeah, there is something we have discussed this afternoon, refer to #1217 . @wuminzhe I've preferred to implement what was discussed in this issue in another pr, so please take a review about this first.

What's the difference between this PR and #1217? readable vs unreadable?

> After discussing with Aki and Echo, We have decided to choose a similar design with solidity, and return the ethabi encode of the result, like precompile_error(module, error_msg)


If they are different and is #1217 approach is better, why we should merge this PR now instead of directly fix #1217 ?

AsceticBear:
This pr just focuses on giving more dispatch error info back to the solidity side, #1217 aims to do some refactor work to make decode precompile error more friendly to developers. I'll not touch this error anymore when in the next step.

HackFisher (approved):

PR #1221, opened 2022-05-18
Follow Ethereum revert message rule

AsceticBear:
The application side gets ethabi error message like the below:
image

Then decode it to readable content:

```
$ ethabi decode params -t string 000000000000000000000000000000000000000000000000000000000000002000000000000000000000000000000000000000000000000000000000000000194e6f7420737570706f72746564206572726f7220696e64657800000000000000
string Not supported error index
```

QA about this: https://ethereum.stackexchange.com/questions/76653/how-to-decode-error-response-from-contact

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1809,11 +1809,13 @@ dependencies = [
  "darwinia-evm",
  "darwinia-evm-precompile-utils",
  "darwinia-support",
+ "ethabi",
  "fp-evm",
  "frame-support",
  "parity-scale-codec 2.3.1",
  "sp-core",
  "sp-io",
+ "sp-std",
 ]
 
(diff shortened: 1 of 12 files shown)


AsceticBear:
@hujw77 @wuminzhe Please take a review

hujw77 (commented):
I think the error message is too long, The message length less than 32 would be better.

AsceticBear:
> I think the error message is too long, The message length less than 32 would be better.

Make sense. I have done some updates, please check it again.

wuminzhe (approved):

hujw77 (approved):
LGTM

HackFisher (approved):

PR #1234, opened 2022-05-26
Add MMR Migration

AurevoirXavier:


frame/header-mmr/src/lib.rs
```diff
@@ -105,10 +105,38 @@ pub mod pallet {
...
 
+	/// Migration step.
+	#[pallet::storage]
+	#[pallet::getter(fn migration_step)]
+	pub type MigrationStep<T> = StorageValue<_, u32, ValueQuery, DefaultMigrationStep>;
+	#[pallet::type_value]
+	pub fn DefaultMigrationStep() -> u32 {
+		500
+	}
+
```
HackFisher, line 116: LGTM, hard code using 500 is simpler?

frame/header-mmr/src/lib.rs
```diff
@@ -105,10 +105,38 @@ pub mod pallet {
...
 	#[pallet::pallet]
 	pub struct Pallet<T>(_);
 	#[pallet::hooks]
 	impl<T: Config> Hooks<BlockNumberFor<T>> for Pallet<T> {
+		fn on_initialize(_n: BlockNumberFor<T>) -> frame_support::weights::Weight {
+			let mut i = 0_u32;
+
+			for _ in
+				frame_support::migration::storage_iter::<u64>(b"DarwiniaHeaderMMR", b"Positions")
+					.drain()
```
HackFisher, line 126: This is iterator, and `drain()` will not loop until last elem, right?

HackFisher (approved):

frame/header-mmr/src/lib.rs
AurevoirXavier: I'm not sure if this will affect the chain performance.

frame/header-mmr/src/lib.rs
AurevoirXavier: Yes. Iterator are lazy-eval.

AsceticBear: Is there any function like `take()`?

AurevoirXavier: `drain` is `take` for map.

frame/header-mmr/src/lib.rs
HackFisher: make sense. We did not have clean the storage right, if we do not set any value.

AurevoirXavier: Yes.

AsceticBear (approved):

frame/header-mmr/src/lib.rs
AsceticBear: I have another about this piece of code. How would you configure it later without a runtime upgrade? I didn't see any setter function. If we can't modify it via signed transaction, has no difference with hard code 500.  @AurevoirXavier

AurevoirXavier: Sudo/Gov

PR #1287, opened 2022-06-16
Clean `darwinia-support` Part.1

AurevoirXavier:


frame/support/src/structs.rs
```diff
@@ -20,62 +20,13 @@
...
-}
-
-#[derive(Clone, PartialEq, Eq, Encode, Decode, RuntimeDebug, MaxEncodedLen, TypeInfo)]
-pub enum LockFor<Balance, Moment> {
-	Common { amount: Balance },
-	Staking(StakingLock<Balance, Moment>),
-}
-
 #[derive(Clone, Default, PartialEq, Eq, Encode, Decode, RuntimeDebug, MaxEncodedLen, TypeInfo)]
 pub struct StakingLock<Balance, Moment> {
```
AsceticBear, line 31: Why not move this to the staking pallet scope?

frame/wormhole/backing/ethereum/src/lib.rs
```diff
@@ -578,10 +578,8 @@ pub mod pallet {
 				if is_ring { <Error<T>>::RingLockedNSBA } else { <Error<T>>::KtonLockedNSBA }
 			);
 			// // Checking redeemer have enough of balance to pay fee, make sure follow up transfer
-			// will success. ensure!(
-			// 	T::RingCurrency::usable_balance(redeemer) >= fee,
-			// 	<Error<T>>::FeeIns
-			// );
+			// will success.
```
AsceticBear, line 581: @xiaoch05 Any suggestion here?

frame/support/src/structs.rs
AurevoirXavier: Still in used.

frame/wormhole/backing/ethereum/src/lib.rs
xiaoch05: Perhaps the fee's system here is not yet fully realized. cc @AurevoirXavier

AurevoirXavier: Yep.
Let's leave it here for now.

AsceticBear (approved):

PR #1300, opened 2022-06-20
Fix RootAccountForPayments

HackFisher:


node/runtime/pangolin/src/pallets/bridge_messages.rs
```diff
@@ -16,7 +16,7 @@ frame_support::parameter_types! {
 	// Shared configurations.
 	pub const MaxMessagesToPruneAtOnce: MessageNonce = 8;
 	// TODO: remove this after FeeMarketPayment upgrade
-	pub RootAccountForPayments: Option<AccountId> = Some(ConcatConverter::<_>::derive_substrate_address((&b"root"[..]).derive_ethereum_address()));
+	pub RootAccountForPayments: Option<AccountId> = Some(ConcatConverter::<_>::derive_substrate_address(H160::from_slice(&b"root0000000000000000"[..])));
```
HackFisher, line 19: This writing looks not right. @AurevoirXavier

HackFisher: > The result is `0x726f6f74303030303030`.

So short? One byte is two hex

AurevoirXavier: Yes. Just a part of it.

But I want to show you this is not the right way to pad zero.

AurevoirXavier (approved):

node/runtime/pangolin/src/pallets/bridge_messages.rs
AurevoirXavier: LGTM.

frame/wormhole/issuing/s2s/src/lib.rs
```diff
@@ -175,7 +177,8 @@ pub mod pallet {
 				&user,
 			)?;
 
-			let backing_address = user.encode().as_slice().derive_ethereum_address();
+			let backing_address = sp_runtime::AccountId32::try_from(
```
AsceticBear, line 180: Why not `use sp_runtime::AccountId32` at the top?

AurevoirXavier (approved):

frame/wormhole/issuing/s2s/src/lib.rs
HackFisher: Just used once, so LGTM

PR #1313, opened 2022-06-28
Remove ed in `AccountBasic`

AsceticBear:


frame/dvm/ethereum/src/lib.rs
```diff
@@ -386,13 +375,13 @@ pub mod pallet {
 	#[pallet::storage]
 	#[pallet::getter(fn get_ring_remaining_balances)]
 	pub(super) type RemainingRingBalance<T: Config> =
-		StorageMap<_, Blake2_128Concat, T::AccountId, RingBalance<T>, ValueQuery>;
+		StorageMap<_, Blake2_128Concat, T::AccountId, u128, ValueQuery>;
```
AsceticBear, line 378: Please be careful about this when reviewing. I think no migration need here because the type is the same.

AurevoirXavier: Recommend using `Currency::Balance`

AsceticBear: How to do that? I have deleted the `RingCurrency` and `KtonCurrency` in the `Config`.

AurevoirXavier: Never mind, I think no any other projects will use this pallet.

HackFisher (approved):

PR #1318, opened 2022-07-04
Avoid Using Hardcode `ChainId` for Internal Transaction

AsceticBear:
For reviewer:

1. https://github.com/ethereum/EIPs/issues/155
2. https://github.com/rust-blockchain/ethereum/blob/master/src/transaction.rs#L61-L77

### DIFF
--- a/frame/dvm/ethereum/src/lib.rs
+++ b/frame/dvm/ethereum/src/lib.rs
@@ -961,7 +961,7 @@ impl<T: Config> InternalTransactHandler for Pallet<T> {
 	fn internal_transact(target: H160, input: Vec<u8>) -> DispatchResultWithPostInfo {
 		let source = T::PalletId::get().derive_ethereum_address();
 		let nonce = <T as darwinia_evm::Config>::RingAccountBasic::account_basic(&source).nonce;
-		let transaction = internal_transaction(nonce, target, input);
+		let transaction = internal_transaction::<T>(nonce, target, input);
 
 		Self::raw_transact(source, transaction).map(|(reason, used_gas)| match reason {
 			// Only when exit_reason is successful, return Ok(...)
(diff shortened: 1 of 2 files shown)


frame/dvm/ethereum/src/lib.rs
```diff
@@ -1076,9 +1080,9 @@ pub fn internal_transaction(nonce: U256, target: H160, input: Vec<u8>) -> Advanc
...
 			// Reference https://github.com/ethereum/EIPs/issues/155
 			//
-			// But this transaction is sent by darwinia-issuing system from `0x0`
-			// So ignore signature checking, simply set `chain_id` to `1`
-			1 * 2 + 36,
+			// The internal transaction is sent by specific pallets, no signature
+			// validation, just create a valid transaction signature is enough.
+			T::ChainId::get() * 2 + 36,
 			H256::from_slice(&[55u8; 32]),
 			H256::from_slice(&[55u8; 32]),
```
HackFisher, line 1087: Any comment for this magic value ?

```
	pub fn new(v: u64, r: H256, s: H256) -> Option<Self> {
		const LOWER: H256 = H256([
			0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
			0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
			0x00, 0x00, 0x00, 0x01,
		]);
		const UPPER: H256 = H256([
			0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff,
			0xff, 0xfe, 0xba, 0xae, 0xdc, 0xe6, 0xaf, 0x48, 0xa0, 0x3b, 0xbf, 0xd2, 0x5e, 0x8c,
			0xd0, 0x36, 0x41, 0x41,
		]);

		let v = TransactionRecoveryId(v);
		let is_valid = v.standard() <= 1 && r < UPPER && r >= LOWER && s < UPPER && s >= LOWER;

		if is_valid {
			Some(Self { v, r, s })
		} else {
			None
		}
	}
```

HackFisher:
And please fix the tests @AsceticBear

frame/dvm/ethereum/src/lib.rs
AsceticBear: I guess it's just a valid value between `LOWER` and `UPPER`.  No specific reason for this. @xiaoch05  Am I right?

xiaoch05: YES

Issue #1325, opened 2022-07-05
Frontier Upgrade For `darwinia-v0.12.4`


AsceticBear:
## Upgrade List

- [ ] https://github.com/paritytech/frontier/pull/583 (**Not all completed**)
- [x] https://github.com/paritytech/frontier/pull/591
- [x] https://github.com/paritytech/frontier/pull/570
- [x] https://github.com/paritytech/frontier/pull/585
- [x] https://github.com/paritytech/frontier/pull/557
- [x] https://github.com/paritytech/frontier/pull/594
- [x] https://github.com/paritytech/frontier/pull/565
- [x] https://github.com/paritytech/frontier/pull/576
- [x] https://github.com/paritytech/frontier/pull/598
- [x] https://github.com/paritytech/frontier/pull/556
- [x] https://github.com/paritytech/frontier/pull/593
- [x] https://github.com/paritytech/frontier/pull/595
- [x] https://github.com/paritytech/frontier/pull/596
- [x] https://github.com/paritytech/frontier/pull/604
- [ ] https://github.com/paritytech/frontier/pull/605
- [x] https://github.com/paritytech/frontier/pull/607
- [x] https://github.com/paritytech/frontier/pull/597
- [ ] https://github.com/paritytech/frontier/pull/617 (**Not all completed**)
- [x] https://github.com/paritytech/frontier/pull/620
- [x] https://github.com/paritytech/frontier/pull/624
- [x] https://github.com/paritytech/frontier/pull/623
- [x] https://github.com/paritytech/frontier/pull/629
- [x] https://github.com/paritytech/frontier/pull/634
- [x] https://github.com/paritytech/frontier/pull/601
- [ ] https://github.com/paritytech/frontier/pull/638
- [x] https://github.com/paritytech/frontier/pull/618
- [x] https://github.com/paritytech/frontier/pull/627
- [x] https://github.com/paritytech/frontier/pull/626
- [x] https://github.com/paritytech/frontier/pull/630
- [x] https://github.com/paritytech/frontier/pull/632
- [x] https://github.com/paritytech/frontier/pull/641
- [x] https://github.com/paritytech/frontier/pull/763

## Note

1. There is a known bug introduced by https://github.com/paritytech/frontier/pull/638. We skip this pr sync as the `darwinia-ethereum` pallet is maintained by us. I create a track issue https://github.com/darwinia-network/darwinia-common/issues/1312, so we won't forget.

AsceticBear:
[...]

PR #1326, opened 2022-07-05
DVM Upgrade For `polkadot-v0.9.13`

AsceticBear:
Upgrade List https://github.com/darwinia-network/darwinia-common/issues/1325


frame/dvm/evm/src/runner/stack.rs
```diff
@@ -66,50 +67,59 @@ impl<T: Config> Runner<T> {
...
-		let max_fee_per_gas = match max_fee_per_gas {
-			Some(max_fee_per_gas) => {
-				ensure!(max_fee_per_gas >= base_fee, <Error<T>>::GasPriceTooLow);
+		let max_fee_per_gas = match (max_fee_per_gas, is_transactional) {
+			(Some(max_fee_per_gas), _) => {
+				ensure!(max_fee_per_gas >= base_fee, Error::<T>::GasPriceTooLow);
 				max_fee_per_gas
 			},
-			None => Default::default(),
+			// Use default gas price for internal transaction
```
HackFisher, line 75: Why there is an internal transaction related comment in this generalized stack runner?

Request more explanation of this part before merging.

HackFisher (changes requested):
```
	/// Pure read-only call to contract, the sender is pallet dvm account.
	/// NOTE: You should never use raw call for any non-read-only operation, be carefully.
	fn read_only_call(contract: H160, input: Vec<u8>) -> Result<Vec<u8>, DispatchError> {
		sp_io::storage::start_transaction();
		let source = T::PalletId::get().derive_ethereum_address();
		let nonce = darwinia_evm::Pallet::<T>::account_basic(&source).nonce;
		let transaction = internal_transaction::<T>(nonce, contract, input);
		let (_, _, info) = Self::execute(source, &transaction, None)?;
		sp_io::storage::rollback_transaction();
		match info {
			CallOrCreateInfo::Call(info) => match info.exit_reason {
				ExitReason::Succeed(_) => Ok(info.value),
				ExitReason::Error(_) => Err(<Error<T>>::InternalTransactionExitError.into()),
				ExitReason::Revert(_) => Err(<Error<T>>::InternalTransactionRevertError.into()),
				ExitReason::Fatal(_) => Err(<Error<T>>::InternalTransactionFatalError.into()),
			},
			_ => Err(<Error<T>>::ReadyOnlyCall.into()),
		}
	}
```

The possible error returned here should be after `sp_io::storage::rollback_transaction();`
```
let (_, _, info) = Self::execute(source, &transaction, None)?;
```

Not directly return using `?`

HackFisher:
Related https://github.com/darwinia-network/darwinia-common/commit/4bc567bbd9d38ad0185ccaf2301a24dfb09180d4

frame/dvm/evm/src/runner/stack.rs
AsceticBear: For internal transactions, the tx sender is a special one and the gas price or (max_fee_per_gas, max_priority_fee_per_gas) is `None` https://github.com/darwinia-network/darwinia-common/blob/main/frame/dvm/ethereum/src/lib.rs#L821-L822. Besides, the internal transactions have the ability to change runtime storage, that means `is_transactional` is true in this case.

But, this will be seen as an invalid call in the frontier upstream runner validation, the `max_fee_per_gas` must be `Some` when `is_transactional` is true. Only if in the non-transactional scenario(estimate gas), the `max_fee_per_gas` can be `None`.

HackFisher: I see, we proceed to review this again after s2s pallet get removed then.

node/service/src/chain_spec/pangoro.rs
```diff
@@ -94,7 +94,9 @@ pub fn genesis_config() -> ChainSpec {
 					array_bytes::hex_into_unchecked(account),
 					GenesisAccount {
 						balance: (MANY_COINS * (10 as Balance).pow(9)).into(),
-						..Default::default()
```
AurevoirXavier, line 97: why?

AsceticBear: Compared with [GenesisAccount Common](https://github.com/darwinia-network/darwinia-common/blob/main/frame/dvm/evm/src/lib.rs#L559), the [GenesisAccount Frontier](https://github.com/darwinia-network/frontier/blob/darwinia-v0.12.4/primitives/evm/src/lib.rs#L66-L77) doesn't have default implementation.

AurevoirXavier (approved):

frame/dvm/evm/src/runner/stack.rs
```diff
@@ -66,50 +67,62 @@ impl<T: Config> Runner<T> {
...
 		let total_payment = value.checked_add(total_fee).ok_or(Error::<T>::PaymentOverflow)?;
 		let source_account = Pallet::<T>::account_basic(&source);
-		ensure!(source_account.balance >= total_payment, <Error<T>>::BalanceLow);
+		// Account balance check is skipped if fee is Zero.
+		// This case is previously verified to only happen on either:
+		// 	- Non-transactional calls.
+		// 	- BaseFee is configured to be Zero.
+		if total_fee > U256::zero() {
+			ensure!(source_account.balance >= total_payment, Error::<T>::BalanceLow);
+		}
```
HackFisher, line 101: Should we add a `TODO` here, supposed to be removed after `internal_transact` be removed.

BTW, why we need this check? It looks dangerous cause possibilities of bypass by accident.

HackFisher: For example, some internal transact with non-zero `msg.value`, but `source_account.balance` is zero?

AsceticBear: > Should we add a TODO here, supposed to be removed after internal_transact be removed.

The validation logic here is consistent with that upstream as much as possible. 

> BTW, why we need this check? It looks dangerous cause possibilities of bypass by accident.

The upstream evm pallet is designed to be integrated independently into the runtime. Obviously, there is some overlap between the checks here and in the ethereum pallet, but these checks would be required if there were no ethereum pallet.

> For example, some internal transact with non-zero msg.value, but source_account.balance is zero?

The internal transaction only accepts input filed from outside, and the value is always zero now. https://github.com/darwinia-network/darwinia-common/blob/main/frame/dvm/ethereum/src/lib.rs#L1066

HackFisher: Could you provide the link to the upstream implementation?

HackFisher: I mean this check `if total_fee > U256::zero()`

AsceticBear: https://github.com/darwinia-network/frontier/blob/darwinia-v0.12.4/frame/evm/src/runner/stack.rs#L100-L105

HackFisher: OK, but the upstream check `if total_fee > U256::zero()` looks not good and not necessary to me.

HackFisher (approved):

PR #1332, opened 2022-07-11
Add `total_supply` function

AsceticBear:
The Kton Erc20 rust needs to know the total supply of the native token.  This pr pave the way for that.

1. Move `account_basic` to a `darwinia-evm` Pallet function.
2. Rename the original `AccountBasic` to `BalanceAdapte`, this trait has one purpose after this refactor, which is handling decimal between the native and evm tokens.
3. Add `evm_total_supply` function


frame/dvm/evm/src/lib.rs
```diff
@@ -472,19 +484,27 @@ pub trait EnsureAddressOrigin<OuterOrigin> {
...
-	/// Get the account basic in EVM format.
-	fn account_basic(address: &H160) -> Account;
-	/// Mutate the basic account.
-	fn mutate_account_basic_balance(address: &H160, new_balance: U256);
-	/// Transfer value.
-	fn transfer(source: &T::AccountId, target: &T::AccountId, value: U256)
-		-> Result<(), ExitError>;
-	/// Get account balance.
+/// A trait for handling balance decimal difference between native and evm tokens.
+pub trait BalanceAdapt<T: Config> {
```
AurevoirXavier, line 488: CurrencyAdapter sounds better.

AurevoirXavier (approved):

frame/dvm/evm/src/lib.rs
AsceticBear: Sounds better. Let me rename it.

AsceticBear: done

PR #1344, opened 2022-07-18
Add BSC Migration

AurevoirXavier:
Which was removed accidentally in #1329.


AsceticBear:
> I found the Bsc pallet is still on the chain.

Why? Doesn't that migration has executed in the last runtime upgrade in the Pangoro network?

AurevoirXavier:
> > I found the Bsc pallet is still on the chain.
> 
> Why? Doesn't that migration has executed in the last runtime upgrade in the Pangoro network?

```diff
-   spec_version: 28140,
+   spec_version: 28160,
```

PR #1363, opened 2022-07-29
Release Tracing Node

AurevoirXavier:


.github/workflows/release.yml
```diff
@@ -30,19 +30,63 @@ jobs:
...
+          mv drml-tracing-node.tar.zst artifacts
+      - name: Upload
+        uses: actions/upload-artifact@v2.2.4
+        with:
+          name: artifacts
+          path: artifacts
+
+  release:
+    name: Release
+    runs-on: ubuntu-latest
```
fewensa, line 75: Missing `needs` for this job.

```
needs: [build, build-with-evm-tracing]
```

AurevoirXavier: Damn.
I have thought about this at the beginning.
But I forgot that at the end.

PR #1373, opened 2022-08-01
Use `NextAuthorities` Instead

AurevoirXavier:
Make things more friendly to bridger.
Bridger only needs to focus on `Authorities` storage while the authority set is changing.

cc @fewensa @hujw77


hujw77 (approved):
LGTM

PR #1387, opened 2022-08-09
Custom KTON Metadata

AsceticBear:


frame/dvm/evm/precompiles/kton/src/lib.rs
```diff
@@ -40,17 +40,25 @@ use frame_support::{
 };
 use sp_core::{H160, U256};
 
-const TOKEN_NAME: &str = "Wrapped KTON";
-const TOKEN_SYMBOL: &str = "WKTON";
-const TOKEN_DECIMAL: u8 = 18;
+/// Metadata of an ERC20 token.
+pub trait ERC20Metadata {
```
AurevoirXavier, line 44: Recommend using `Erc20Metadata`

AsceticBear: I was torn between those two several times. :smile:

frame/dvm/evm/precompiles/kton/src/lib.rs
```diff
@@ -40,17 +40,25 @@ use frame_support::{
 };
 use sp_core::{H160, U256};
 
-const TOKEN_NAME: &str = "Wrapped KTON";
-const TOKEN_SYMBOL: &str = "WKTON";
-const TOKEN_DECIMAL: u8 = 18;
+/// Metadata of an ERC20 token.
+pub trait Erc20Metadata {
```
AurevoirXavier, line 44: Should we put this in kton precompile?
It looks like a more generic property.

AsceticBear: Let's keep it here for now, no other modules will be used it at the moment.

HackFisher (approved):

Issue #1403, opened 2022-08-17
Clean staled order data


AsceticBear:
In the `pangoroFeeMarket` of the Pangolin network, there are some staled old orders that will never be confirmed in the future.
But it reduce the relayer's order capacity, we need to remove it by migration.

```sh
2022-08-17 15:41:43.830  WARN tokio-runtime-worker pallet_fee_market: bear: --- order Order { lane: [0, 0, 0, 0], message: 11, sent_time: 2320090, confirm_time: None, locked_collateral: 50000000000, assigned_relayers: [AssignedRelayer { id: f41d3260d736f5b3db8a6351766e97619ea35972546a5f850bbf0b27764abe03 (5HanDP7Y...), fee: 15000000000, valid_range: 2320090..2320690 }, AssignedRelayer { id: d43593c715fdd31c61141abd04a99fd6822c8558854ccde39a5684e7a56da27d (5GrwvaEF...), fee: 15000000000, valid_range: 2320690..2321290 }, AssignedRelayer { id: a09c083ca783d2f2621ae7e2ee8d285c8cf103303f309b031521967db57bda14 (5FhHsQmX...), fee: 20000000000, valid_range: 2321290..2321890 }] }    
2022-08-17 15:41:43.830  WARN tokio-runtime-worker pallet_fee_market: bear: --- order Order { lane: [0, 0, 0, 0], message: 1, sent_time: 2133269, confirm_time: None, locked_collateral: 50000000000, assigned_relayers: [AssignedRelayer { id: f41d3260d736f5b3db8a6351766e97619ea35972546a5f850bbf0b27764abe03 (5HanDP7Y...), fee: 15000000000, valid_range: 2133269..2133869 }, AssignedRelayer { id: d43593c715fdd31c61141abd04a99fd6822c8558854ccde39a5684e7a56da27d (5GrwvaEF...), fee: 15000000000, valid_range: 2133869..2134469 }, AssignedRelayer { id: a09c083ca783d2f2621ae7e2ee8d285c8cf103303f309b031521967db57bda14 (5FhHsQmX...), fee: 20000000000, valid_range: 2134469..2135069 }] }    
[...]

PR #1404, opened 2022-08-17
Only keep the lane monitored by the bridgers

AsceticBear:
Part of #1403. If someone sends cross-chain messages to those unmonitored lanes, nothing will be seen in the target chain.

Bridger Lane Config Reference: https://github.com/darwinia-network/bridger/blob/master/.maintain/config/bridge-pangolin-pangolinparachain.toml


AurevoirXavier (approved):
Require a companion.

AsceticBear:
> Need a companion.

Why? This will not influence the Darwinia/Crab network.

PR #1426, opened 2022-09-06
Release `v2.9.4`

AurevoirXavier:


AurevoirXavier:
@AsceticBear Please help check the EVM mapping worker `sync_from`.

AsceticBear:
Better to include #1425 in this release

AsceticBear:
> Please help check the EVM mapping worker `sync_from`.

LGTM, It's okay for me if we only reset the pangolin network. How to deal with the existing storage related to the pangoro network?

HackFisher (approved):

AurevoirXavier:
The genesis is generated before #1425. 

But both spec versions are 29040.

If we release based on this commit, there will be two different runtime logic.

Issue #1239, opened 2022-05-30
Remove wormhole/helix pallets, and re-implement using message sdk.


HackFisher:
- [ ] Fix https://github.com/darwinia-network/darwinia-common/issues/1215 and replace it with generalized call back later.
- [x] Remove wormhole/helix pallets: https://github.com/darwinia-network/darwinia-common/tree/main/frame/wormhole
- [x] Re-implement using message sdk. Related #1235 #1227 
- [x] Clean and generalize `internal_call` and `read_only_call`, we should move source as parameter, merge with `internal_transact_with_source_account ` @AsceticBear

AurevoirXavier:
Please update this.

Issue #1157, opened 2022-04-02
Revamp `darwinia-support`


AurevoirXavier:
Too many unrelated items within the same context.
Maybe we can split them into multiple specific crates. But leave the common things.

AurevoirXavier:
Do not clean the code now.
Draft a list first, that would be better.
@AsceticBear

PR #1436, opened 2022-09-09
Delete s2s related pallets

AsceticBear:
Part of https://github.com/darwinia-network/darwinia-common/issues/1239.

Get rid of codes related to the backing and issuing S2S. As for the internal transaction part, I prefer to raise another PR later so that it is clearer.


HackFisher (approved):

AurevoirXavier:
Part of #1157.

Issue #749, opened 2021-08-16
Magic TransactionSignature in internal transaction


HackFisher:
https://github.com/darwinia-network/darwinia-common/pull/713/files#diff-400a8a292c291b6925055546f3901212cbea818b0c7b9806a1f7eb18290160a9R82

Why the parameters and chain id are hardcoded, does it still apply to testnet?

AsceticBear:
> Why the parameters and chain id are hardcoded, does it still apply to testnet?

Would you like to provide more detailed info about this?

HackFisher:
Following part in https://github.com/darwinia-network/darwinia-common/blob/00cc6ea6753bdb9d3750a9790a73ff6eafd36613/frame/dvm/rpc/runtime-api/src/lib.rs#L68 still looks magic to me:
```
signature: ethereum::TransactionSignature::new(
				// Reference https://github.com/ethereum/EIPs/issues/155
				//
				// But this transaction is sent by darwinia-issuing system from `0x0`
				// So ignore signature checking, simply set `chain_id` to `1`
				1 * 2 + 36,
				H256::from_slice(&[55u8; 32]),
				H256::from_slice(&[55u8; 32]),
			)
```

AsceticBear:
The signature `new_internal_transaction` is just mock data that will not be used in later contexts. It would be much clear if there is a default implementation for TransactionSignature. check https://github.com/rust-blockchain/ethereum/pull/18

AsceticBear:
Once https://github.com/paritytech/frontier/pull/481 is accepted, I think we can remove this black magic signature.

AsceticBear:
This can be close after s2s-related pallets are removed.

Issue #1385, opened 2022-08-09
Review internal transact related code in stack runner


HackFisher:
For internal transactions, the tx sender is a special one and the gas price or (max_fee_per_gas, max_priority_fee_per_gas) is `None` https://github.com/darwinia-network/darwinia-common/blob/main/frame/dvm/ethereum/src/lib.rs#L821-L822. Besides, the internal transactions have the ability to change runtime storage, that means `is_transactional` is true in this case.

But, this will be seen as an invalid call in the frontier upstream runner validation, the `max_fee_per_gas` must be `Some` when `is_transactional` is true. Only if in the non-transactional scenario(estimate gas), the `max_fee_per_gas` can be `None`.

Maybe I should add more notes here about this validation branch. Another thing is that the internal transaction is no longer needed as soon as the s2s pallets are removed in this repo. It's in processing now.

_Originally posted by @AsceticBear in https://github.com/darwinia-network/darwinia-common/pull/1326#discussion_r940253254_

PR #1441, opened 2022-09-15
Remove `internal_transaction`

AsceticBear:
Close https://github.com/darwinia-network/darwinia-common/issues/749 https://github.com/darwinia-network/darwinia-common/issues/1385

1. delete deprecated `darwinia-evm-precompile-transfer` precompile
2. Remove `internal_transaction` part and add unite tests to ensure the `MessageRootGetter` functionality is not broken.


frame/dvm/ethereum/src/lib.rs
```diff
@@ -927,63 +880,6 @@ impl<T: Config> Pallet<T> {
...
-	/// Read-only call to deployed evm contracts.
-	fn read_only_call(contract: H160, input: Vec<u8>) -> Result<Vec<u8>, DispatchError>;
-}
-
-impl<T: Config> InternalTransactHandler for Pallet<T> {
-	/// Execute transaction from other pallets(internal transaction)
-	/// NOTE: The difference between the rpc transaction and the internal transaction is that
-	/// The internal transactions will catch and throw evm error comes from runner to caller.
-	fn internal_transact(target: H160, input: Vec<u8>) -> DispatchResultWithPostInfo {
-		let source = T::PalletId::get().derive_ethereum_address();
```
HackFisher, line 943: @hujw77 HelixDAO might have some token migration for this? The allocated part from Darwinia to Crab as xRING?

@xiaoch05

frame/bridge/message-gadget/src/lib.rs
```diff
@@ -76,30 +80,35 @@ const LOG_TARGET: &str = "runtime::message-gadget";
...
 			<CommitmentContract<T>>::get(),
 			hashing::keccak_256(b"commitment()")[..4].to_vec(),
+			0.into(),
+			1_000_000_000_000,
+			None,
+			None,
+			None,
+			vec![],
+			false, // is_transactional = false, use the default gas_price
+			<T as darwinia_evm::Config>::config(),
```
HackFisher, line 97: @AurevoirXavier Request further review

frame/dvm/ethereum/src/lib.rs
xiaoch05: we need this internal_transact and root_transact to migrate xRING from a pallet to a ethereum account.
see https://github.com/darwinia-network/darwinia/issues/726

AsceticBear: Can we do this operation on the Crab network before this pr has been upgraded to the Crab runtime? Because the internal_transact is still available in the current runtime(on-chain).

HackFisher: Yes, I recommend to do that before runtime upgrade

HackFisher (approved):

PR #1449, opened 2022-09-22
Add LCMP message validation event

### LINKED ISSUE #1440: Expose `message_transact`  validation reason

AsceticBear:
Fix #1440 

@wuminzhe Please pay attention to this pr. This event makes debugging easier on the SDK side.


frame/dvm/ethereum/src/lib.rs
```diff
@@ -285,21 +286,21 @@ pub mod pallet {
...
-				_ => Err(Error::<T>::MessageTransactionError),
+				_ => Err(Error::<T>::InvalidTransactionType),
 			}?;
 
-			ensure!(
-				Self::validate_transaction_in_block(source, &extracted_transaction).is_ok(),
-				Error::<T>::MessageValidateError
-			);
+			if let Err(err) = Self::validate_transaction_in_block(source, &extracted_transaction) {
+				Self::deposit_event(Event::MessageTransactValidateRejected { source, nonce, err });
```
HackFisher, line 301: Can't we tell The reason using the Error since Error is also enum?

Could we add struct to MessageValidateRejected.

This way of adding event looks weird.

HackFisher (changes requested):
Didn't see cases of Event for error/failure purpose report

AsceticBear:
> Didn't see cases of using Event for error/failure purpose report

There are some cases both in the substrate or parity-bridges-common, like:

- https://github.com/darwinia-network/darwinia-messages-substrate/blob/main/modules/dispatch/src/lib.rs#L203
- https://github.com/darwinia-network/darwinia-messages-substrate/blob/main/modules/dispatch/src/lib.rs#L228
- https://github.com/paritytech/substrate/blob/master/frame/referenda/src/lib.rs#L1026

frame/dvm/ethereum/src/lib.rs
AsceticBear: > Can't we tell The reason using the Error since Error is also enum?
> Could we add struct to MessageValidateRejected.

As for as I know, the `Pallet Error` does not support struct type now, so the `MessageValidateRejected` can not contain the more detailed message. Another way is to create many runtime error types, then needs to convert the validated error type to the corresponding runtime error type.

HackFisher:
> > Didn't see cases of using Event for error/failure purpose report
> 
> There are some cases both in the substrate or parity-bridges-common, like:
> 
> * https://github.com/darwinia-network/darwinia-messages-substrate/blob/main/modules/dispatch/src/lib.rs#L203
> * https://github.com/darwinia-network/darwinia-messages-substrate/blob/main/modules/dispatch/src/lib.rs#L228
> * https://github.com/paritytech/substrate/blob/master/frame/referenda/src/lib.rs#L1026

There are still some difference there, they are returning result instead of dispatch result with Error.

Would rather to make it pending and wait for the substrate update if it is not high priority.

AsceticBear:
Wait for https://github.com/paritytech/substrate/pull/10242

AsceticBear:
This will be solved more properly in darwinia-2.0, Close it now.

PR #1450, opened 2022-09-23
Code clean

AsceticBear:
Too many `*` make it hard to tell where a particular variable, trait, or const comes from. And it is hard to tell which is useless in the upstream repo. Related to https://github.com/darwinia-network/darwinia-messages-substrate/pull/189


AurevoirXavier:
I think the upstream should not take care of how the downstream use it.

HackFisher (approved):
LGTM

AurevoirXavier:
Need a companion for Darwinia.

AsceticBear:
> Need a companion for Darwinia.

Do it together with other companions in the next Darwinia release?

